PR #3, opened 2026-01-26
Filter self-messages in the Discord module to prevent reingestion and cascading identity issues

Anarchid:


Anarchid:
included in #4

PR #26, opened 2026-04-23
Assorted conhost fixes: MCPL policy, stderr forwarding, typing metadata, workspace watcher

Anarchid:
## Summary

Batch of small conhost-driven improvements discovered while integrating AF into connectome-host. Each commit stands alone; ordered by topic rather than dependency.

**MCPL**
- `feat(mcpl): per-server tool allow/deny policy` — `enabledTools`/`disabledTools` on `McplServerConfig`, bare names with `*` wildcards. Filtered both at `tools/list` time and at dispatch time — schema omission alone doesn't stop a model from imitating a denied call it's already seen in history. Deny wins on overlap.
- `feat(mcpl): forward subprocess stderr as mcpl:server-stderr trace events` — child stderr was previously silently dropped. Now every line becomes a trace event keyed on `serverId`, with pre-handshake lines buffered and drained once the instance exists; rebindable for the reconnect probe → persistent swap.
- `feat(mcpl): thread opaque routing metadata through channels/typing` — `sendChannelsTyping` / `ChannelRegistry.startTyping` accept an optional `Record<string, unknown>` that travels with the notification. Intentionally untyped per-server (Zulip reads `topic`, others pick their own keys). Registry caches per-channel metadata so the 7s refresh stays on the same thread.
- `feat(mcpl): explicit stop op on channels/typing` — `op: 'start' | 'stop'` added; `ChannelRegistry.stopTyping` dispatches one final stop before clearing the interval, so servers that support explicit stop (Zulip) clear immediately instead of waiting for auto-expire. Metadata carries on the stop too.

**Framework**
- `feat(framework): expose ChannelRegistry on AgentFramework` — was a private field; modules that need channel-level ops (typing, default publish channel) had no way in. Added a `channels` getter (null when no MCPL servers) and re-exported `ChannelRegistry` from the package entry so consumers can type against it.

**Workspace**
- `feat(workspace): public resolveAbsolutePath for peer modules` — turns a `mount/sub/file.md` path into an absolute FS path without peer modules reaching into private `parsePath` state. Honors the same path-traversal guard.
- `fix(workspace): chokidar-correct-use mitigations for silent watcher failures` — reproducibly on Linux/WSL, chokidar 4's fs.watch backend silently drops events when the watched path (or any parent) doesn't exist at subscribe time and surfaces no error without an `error` handler. Post-failure state is indistinguishable from watching an empty dir. Three mitigations, all canonical chokidar usage:
  - `mkdir -p` the watch path before attach for read-write mounts (read-only left alone).
  - `WatcherLifecycle` callbacks (`onReady`, `onError`) forwarded as `workspace:watcher-error` events.
  - Mount state carries `watcherReadyAt` + `watcherError`, persisted via snapshot. Null `watcherReadyAt` on a live session's mount = watcher never attached; set timestamp with empty tree = attached-but-empty. `watcherReadyAt` is intentionally not restored across restarts — stale timestamp would mask a fresh attach failure.
  - Does not address inode-replace (dir rm'd + recreated post-subscribe); that needs re-attach logic / `@parcel/watcher`.

## Test plan
- [x] `npx tsc --noEmit` clean
- [ ] `node --import tsx --test test/tool-policy.test.ts`
- [ ] `node --import tsx --test test/workspace-watcher.test.ts`
- [ ] Smoke test in connectome-host: zulip typing indicators clear promptly, disabled tool denied both at list and dispatch, mcpl server stderr surfaces as traces

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -51,6 +51,7 @@ import { InferenceRouter } from './mcpl/inference-router.js';
 import { ChannelRegistry } from './mcpl/channel-registry.js';
 import { safeSlice } from './safe-slice.js';
 import { CheckpointManager } from './mcpl/checkpoint-manager.js';
+import { isToolAllowed } from './mcpl/tool-policy.js';
 import { EventGate } from './gate/event-gate.js';
 import { UsageTracker, type PersistedUsageState } from './usage/usage-tracker.js';
 import type { SessionUsageSnapshot, UsageUpdatedEvent } from './usage/types.js';
@@ -2111,6 +2112,13 @@ export class AgentFramework {
     }
     const prefix = config.toolPrefix ?? `mcpl--${config.id}`;
     const toolName = call.name.slice(prefix.length + 2); // Strip prefix + '--'
+    if (!isToolAllowed(toolName, config)) {
+      return {
+        success: false,
+        error: `Tool '${call.name}' is not permitted by this server's tool policy.`,
+        isError: true,
+      };
+    }
     try {
       const result = await server.sendToolsCall(toolName, call.input as Record<string, unknown>);
       return {
@@ -2660,6 +2668,7 @@ export class AgentFramework {
       try {
         const result = await server.sendToolsList();
         for (const tool of result.tools) {
+          if (!isToolAllowed(tool.name, config)) continue;
           // MCP tool schemas are generic JSON Schema; cast to membrane's ToolDefinition format
           const schema = tool.inputSchema as import('./types/index.js').ToolDefinition['inputSchema'];
           tools.push({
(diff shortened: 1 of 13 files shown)


Anarchid:
Follow-up commit [`32cb10c`](https://github.com/Anarchid/agent-framework/commit/32cb10c) addresses internal review findings:

**🟡 stopTyping asymmetry** — single-channel branch now dispatches the `stop` notification only when an interval was actually active, matching the global-clear branch. Defensive `stopTyping(ch)` calls no longer spam `stop` notifications at servers that never saw a `start`.

**🟡 Mid-typing metadata update latency** — `startTyping(ch, newMetadata)` on an already-typing channel now dispatches an immediate refresh when the metadata differs from the cached one, instead of waiting up to 7s for the next tick. This is exactly the Zulip topic-switch case the op:'stop' commit was meant to cover; the fast-path refresh reuses the same routing-hint contract without stop/start flap.

**🟡 Reconnect stderr rebind** — added a code comment acknowledging it inherits the same "transfer internals, leak listeners on the dead instance" pattern as the message-routing/lifecycle rewiring. Not reworked — worth knowing for whoever cleans that reconnect shape up.

New test: `test/channel-registry-typing.test.ts` — 6 cases covering both symmetry and the metadata-change fast-path. Full suite green (88 tests across 7 files).

PR #28, opened 2026-05-07
feat(gate): rate_limit + passive_sample behaviors, metadataTrue match, resets

Tengro:
## Summary

Two new generic `GateBehavior` types for the steady-stream-with-quotas case where existing `always`/`skip`/`debounce` aren't expressive enough:

- **`rate_limit`** — token bucket per `keyBy`-resolved metadata field. Empty bucket → trigger denied.
- **`passive_sample`** — fire every Nth match, then reset the counter. Optional `keyBy` for per-key counters.

Plus three small supporting bits:

- **`match.metadataTrue: string[]`** — OR-ing flag matcher (uses sane truthiness — empty array isn't truthy).
- **`policy.resets: string[]`** — when a policy fires, clear named target policies' rate-limit / passive-sample state. Lets configs express interactions like "a direct invocation resets the passive-sample counter" without coupling behaviors to each other in code.
- **`gate:status`** exposes per-policy `rateLimitState` (`bucketCount`, `deniedCount`) and `passiveSampleState` (`counterCount`, `fireCount`).

Both new behaviors terminate the policy chain with `trigger: false` on deny — no fallthrough to gate-level `default`. Buckets / counters fall back to a single shared partition when `keyBy` is omitted or the metadata field is missing on a given event.

## Why generic, not chat-specific

The behaviors are agnostic to chat semantics. `rate_limit` keyed by any metadata field works for per-tenant API throttling, per-source sensor rate caps, anything bucket-shaped. `passive_sample` works for "decide whether to act every Nth observation" — chat passive sampling, periodic analysis of a sensor stream, etc. The motivating use case is chat-bot triggering, but no chat assumptions are baked in.

## Internal

- Constructor takes optional `now: () => number` for deterministic time in tests; production keeps using `Date.now`.
- `reloadIfChanged` cleans up rate_limit / passive_sample state for policies removed by `gate.json` edits, mirroring how debounce timers are already cleaned up.
- `dispose()` clears the new state maps.

## Example (chat-trigger config)

```json
{
  "policies": [
    {
      "name": "direct-invocation",
      "match": {
        "scope": ["mcpl:channel-incoming"],
        "metadataTrue": ["isMention", "isPrivate", "isReplyToBot"]
      },
      "behavior": {
        "rate_limit": { "tokens": 12, "refillIntervalMs": 7200000, "keyBy": "senderId" }
      },
      "resets": ["passive-sample"]
    },
    {
      "name": "passive-sample",
      "match": { "scope": ["mcpl:channel-incoming"] },
      "behavior": {
        "passive_sample": { "every": 100, "keyBy": "channelId" }
      }
    }
  ],
  "default": "skip"
}
```

## Test plan
- [x] 21 new tests in `test/event-gate-rate-limit.test.ts` cover validation reporting, `metadataTrue` (truthy / empty-array / empty-string), `rate_limit` (drain / refill / cap / keyBy isolation / missing-key shared bucket / no-fallthrough), `passive_sample` (fire / reset / keyBy isolation / no-fallthrough), `resets` (fires-clear-targets, denies-do-not-reset, always+resets, unknown targets ignored), and status output.
- [x] Full existing 107-test suite passes unchanged. Total: 128 / 128.
- [x] `npx tsc --noEmit` clean.

### DIFF
--- a/src/gate/event-gate.ts
+++ b/src/gate/event-gate.ts
@@ -147,6 +147,17 @@ function formatBehavior(b: import('./types.js').GateBehavior): string {
   return 'unknown';
 }
 
+/**
+ * Stable fingerprint for a GateBehavior, used during reload to detect
+ * config changes that invalidate per-policy state. JSON.stringify is
+ * deterministic for our small, flat behavior shapes — keys appear in
+ * declaration order — so this is enough to spot any change in tokens,
+ * refillIntervalMs, keyBy, every, or debounce duration.
+ */
+function behaviorFingerprint(b: import('./types.js').GateBehavior): string {
+  return typeof b === 'string' ? b : JSON.stringify(b);
+}
+
 // ---------------------------------------------------------------------------
 // Config validation
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Tengro:
Addressed the reload-stale-state concern in 8898da2.

## What changed

- New `clearPolicyState(name, options)` helper that wipes debounce timer + rate-limit bucket + passive-sample counter (preserves historical stats).
- New `behaviorFingerprint(behavior)` hash via `JSON.stringify` — deterministic for the flat behavior shapes we have.
- `reloadIfChanged` now compares old vs new behavior hashes per kept policy. When the hash differs, state is cleared so the new params take effect on the very next evaluate().
- Removed-policy cleanup also routes through the same helper (less duplication than the original PR).
- `gate:config-reloaded` trace event gains optional `removedPolicies` and `stateClearedPolicies` so operators can see exactly which edits invalidated state.

## Tests added (8 new)

| Edit | Expectation |
|---|---|
| `tokens` lowered 12 → 3 | Fresh 3-token bucket at the new limit |
| `refillIntervalMs` changed | Bucket cleared, fresh tokens |
| `keyBy: "channelId"` → `"senderId"` | Old channel-keyed bucket gone |
| `passive_sample.every` 10 → 3 | Counter cleared, doesn't fire stale (counter at 7+1=8 would have fired ≥ 3 without the fix) |
| `passive_sample.keyBy` changed | Per-key counters cleared |
| Identical behavior on reload | State preserved (negative test) |
| Trace event content | `stateClearedPolicies` lists changed names |
| Trace event content | `removedPolicies` separate from state-cleared |

Full suite: **136 / 136 passing**, `npx tsc --noEmit` clean.

PR #35, opened 2026-05-21
feat(trace): surface content-block lifecycle + per-chunk block type

Anarchid:
## Summary

`driveStream` was collapsing every stream event with `content` into a flat `inference:tokens` trace event with just `agentName + content`. The membrane already attaches block-type metadata to each chunk (`ChunkMeta.type`, `.blockIndex`) and emits `'block'` lifecycle events on the yielding stream — but the framework discarded both, making it impossible for downstream consumers (TUIs, WebUIs) to render thinking content distinctly from regular text.

This threads what the membrane already provides:

- `inference:tokens` gains optional `blockType` and `blockIndex` fields, copied verbatim from the chunk's `ChunkMeta`.
- New `inference:content_block` trace event, emitted on every `block_start` / `block_complete` from the membrane. Consumers switch render lanes on `block_start` (e.g. open a "thinking" element when `blockType === 'thinking'`).

## Dependency

Pairs with antra-tess/membrane#19, which fixes the native-tools yielding path that was hardcoding `meta.type = 'text'` and skipping block events entirely. Without that fix this plumbing is correct but useless; with it, both paths (prefill and native) now hand the framework consistent block-aware events.

## Backward compatibility

Purely additive: existing trace consumers continue to receive `inference:tokens` with `content`; the new fields are optional and the new event variant is opt-in.

## Test plan

- [x] \`npm run build\` (tsc) — clean
- [x] \`node --import tsx --test test/framework.test.ts\` — 19/19 pass
- [x] End-to-end against conhost (separate PR) confirms thinking lane now renders correctly when paired with the membrane fix.

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -1827,20 +1827,22 @@ export class AgentFramework {
               type: 'inference:tokens',
               agentName: agent.name,
               content: event.content,
-              blockType: event.meta?.type,
-              blockIndex: event.meta?.blockIndex,
+              blockType: event.meta.type,
+              blockIndex: event.meta.blockIndex,
             });
             break;
 
-          case 'block':
+          case 'block': {
+            const { event: phase, index, block } = event.event;
             this.emitTrace({
               type: 'inference:content_block',
               agentName: agent.name,
-              event: event.event.event,
-              blockType: event.event.block.type,
-              blockIndex: event.event.index,
+              phase,
+              blockType: block.type,
+              blockIndex: index,
             });
             break;
+          }
 
           case 'tool-calls': {
             hadToolCalls = true;
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -1827,6 +1827,18 @@ export class AgentFramework {
               type: 'inference:tokens',
               agentName: agent.name,
               content: event.content,
+              blockType: event.meta?.type,
+              blockIndex: event.meta?.blockIndex,
+            });
+            break;
+
+          case 'block':
+            this.emitTrace({
+              type: 'inference:content_block',
+              agentName: agent.name,
+              event: event.event.event,
+              blockType: event.event.block.type,
+              blockIndex: event.event.index,
             });
             break;
 
--- a/src/types/trace.ts
+++ b/src/types/trace.ts
@@ -67,6 +67,27 @@ export type TraceEvent =
       type: 'inference:tokens';
       agentName: string;
       content: string;
+      /**
+       * Which membrane block produced this chunk. Carried verbatim from the
+       * membrane's ChunkMeta. Consumers that want to render thinking content
+       * distinctly from regular text should switch on this.
+       */
+      blockType?: 'text' | 'thinking' | 'tool_call' | 'tool_result';
+      /** 0-indexed block position in the current assistant turn. */
+      blockIndex?: number;
+    })
+  | (TraceEventBase & {
+      /**
+       * Structural boundary inside the assistant turn — emitted on every
+       * block_start / block_complete from the membrane. Use this to switch
+       * render lanes (e.g. open a dim "thinking" element on block_start
+       * with blockType='thinking', close it on block_complete).
+       */
+      type: 'inference:content_block';
+      agentName: string;
+      event: 'block_start' | 'block_complete';
+      blockType: 'text' | 'thinking' | 'tool_call' | 'tool_result';
+      blockIndex: number;
     })
   | (TraceEventBase & {
       type: 'inference:tool_calls_yielded';
--- a/src/types/trace.ts
+++ b/src/types/trace.ts
@@ -72,9 +72,9 @@ export type TraceEvent =
        * membrane's ChunkMeta. Consumers that want to render thinking content
        * distinctly from regular text should switch on this.
        */
-      blockType?: 'text' | 'thinking' | 'tool_call' | 'tool_result';
+      blockType: 'text' | 'thinking' | 'tool_call' | 'tool_result';
       /** 0-indexed block position in the current assistant turn. */
-      blockIndex?: number;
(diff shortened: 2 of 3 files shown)


Anarchid:
# 🟡 MEH

PR 35 is *technically fine* in the way a tax return is technically fine — nothing the auditor can throw you in jail for, but you also didn't try. `tsc` is clean, 19/19 tests pass, the diff is 33 lines that do approximately what the description promises. And yet.

## Findings

### 🟠 CONCERNING — Optional-chain theatre forces nullability onto every downstream consumer

**`src/framework.ts:1830-1831`**
```ts
blockType: event.meta?.type,
blockIndex: event.meta?.blockIndex,
```

The PR description correctly notes that "the membrane already attaches block-type metadata to each chunk." `TokensEvent.meta` is `ChunkMeta` — **non-optional**, with `type` and `blockIndex` as **required** fields. Verified against the installed `@animalabs/membrane@0.5.48`. There is no universe in which `event.meta` is `undefined` at this call site.

But the `?.` is there anyway, like reflexively crossing oneself at a funeral. That paranoia leaks into the type:

**`src/types/trace.ts:71-75`**
```ts
blockType?: 'text' | 'thinking' | 'tool_call' | 'tool_result';
blockIndex?: number;
```

Now every TUI and WebUI consumer this PR was written for has to write `if (event.blockType) { ... }` forever, branching on a case that **cannot occur**. The whole *point* was to let consumers switch render lanes — the switch is now nullable for no reason.

**Fix:** drop the optional chains, make the fields required:
```ts
blockType: event.meta.type,
blockIndex: event.meta.blockIndex,
```
```ts
blockType: 'text' | 'thinking' | 'tool_call' | 'tool_result';
blockIndex: number;
```
If older membranes shipping partial `meta` is a real worry, raise the peer-dep floor — don't apologize for it at every call site.

### 🟡 MEH — No test asserts the new payload exists

**`test/framework.test.ts:627`**
```ts
assert.ok(events.includes('inference:tokens'));
```

This is the totality of "test coverage of the new fields." It checks that the string `'inference:tokens'` appeared in an array. It does **not** check that the new `blockType` / `blockIndex` are populated. The new `'inference:content_block'` variant has *zero* tests that emit a `'block'` stream event to exercise the new switch arm. The mock stream at `test/framework.test.ts:60-64` never emits `{ type: 'block', ... }`.

This is a PR whose entire purpose is "downstream consumers can read these new fields," and there is no `assert.strictEqual(tokenEvent.blockType, 'text')` anywhere in the suite. The "19/19 pass" in the test plan is technically true and ethically empty. The new code paths are uncovered.

**Fix:** add (a) an assertion on the `blockType`/`blockIndex` fields of a captured `inference:tokens` event, and (b) a mock-stream test that pushes a `{ type: 'block', event: { event: 'block_start', index: 0, block: { type: 'thinking' } } }` and asserts the resulting `inference:content_block` trace.

### 🟡 MEH — `event.event.event`

**`src/framework.ts:1839`**
```ts
event: event.event.event,
```

This is what happens when nobody owns the naming of the type hierarchy. `StreamEvent` has `{ type: 'block', event: BlockEvent }`, and `BlockEvent` has its own `{ event: 'block_start' | 'block_complete', ... }`. So in the framework one destructures `event.event.event` and then writes `event:` again on the trace payload, which is its *own* collision because the trace type already has a `type:` discriminator and now also an `event:` sub-discriminator. Reading this case-arm is like watching a fish discover water.

Most of the fault is upstream in `membrane/src/types/streaming.ts` (`BlockEvent.event` is the original sin). But this PR accepted the shape uncritically and propagated the bad name onto the trace payload.

**Fix:** rename the trace-payload field to `phase` (or `lifecycle`) so consumers don't have to write `if (traceEvent.event === 'block_start')` — that line reads like a stutter. And destructure once at the top of the case arm:
```ts
case 'block': {
  const { event: phase, index, block } = event.event;
  this.emitTrace({
    type: 'inference:content_block',
    agentName: agent.name,
    phase,
    blockType: block.type,
    blockIndex: index,
  });
  break;
}
```

## Tooling

```
$ npx tsc --noEmit
EXIT: 0

$ node --import tsx --test test/framework.test.ts
# tests 19
# suites 4
# pass 19
# fail 0
```

## Verdict

The diff is correct, additive, and small. It does not break the state machine, the module contract, or trace/process event separation. It will work. But it ships with paranoid nullability that contradicts the upstream type contract, zero test assertions on the fields whose existence is the entire reason for the PR, and a `event.event.event` line that future readers will have to puzzle through for no benefit. The membrane-side dependency (antra-tess/membrane#19) is doing the real work; this is the plumbing that bolts onto it, and it should have been tighter.

Tighten the types, write the tests, rename the field. Then it's a 🟢.

---
🤖 Roast generated with [Claude Code](https://claude.com/claude-code) via `/roast`

Anarchid:
Addressed in 2d19aec.

**🟠 Optional-chain theatre** — dropped `?.` in driveStream, `blockType`/`blockIndex` are now required on `inference:tokens`. Verified against `ChunkMeta`: both fields are required there, the optional access was contradicting the upstream contract for no reason.

**🟡 No test asserts the new payload** — added a focused test that captures the trace stream and asserts (a) every `inference:tokens` carries `blockType: 'text'` and `blockIndex: 0` from the mock's `ChunkMeta`, and (b) `inference:content_block` events fire in `block_start` → `block_complete` lifecycle order with the right block metadata. Mock yielding stream now also emits the block lifecycle events around tokens, so the existing tests exercise the new case-arm too.

**🟡 `event.event.event`** — renamed the trace-payload discriminator from `event` to `phase`, destructured at the top of the case-arm. Consumer-side update follows in connectome-host#32 (the only known consumer).

20/20 tests pass, tsc clean.

PR #39, opened 2026-05-30
fix(mcpl): lazy-register inbound channels + surface speech-routing failures

antra-tess:
## Problem
A Discord channel that delivered an incoming message *before* its `channels/register` / `channels/changed` registration landed — or when that event was missed entirely (e.g. the bot gains visibility via a permission change that fires no `channelCreate`) — was usable for **input** (it woke the agent and became `defaultPublishChannel`) but not as an **output** locus. `routeSpeech()` couldn't resolve it in `this.channels`, so the agent's reply was dropped **silently**. Observed live: an explicit @mention got a generated reply that never reached Discord.

## Changes
- **`ChannelRegistry.handleIncoming`** lazy-registers an unknown channel from the inbound message itself (`serverId` + `channelId` + `channelName`, `open: true`), so it is immediately a publishable locus. A later authoritative `channels/register`/`changed` overwrites the stub.
- **`routeSpeech()`** now emits an `mcpl:speech-route-failed` trace and calls a new `onRouteFailure` callback on every failure path. `framework.ts` wires it to record a `[discord-send-failed]` marker into chronicle via `addMessage` (no inference triggered → no wake loop), so a dropped reply is visible to the agent and operator instead of vanishing.
- Fixed `routeSpeech` returning `{delivered:true}` unconditionally — it now returns the real `delivered` value and treats a server-reported `delivered:false` as a failure.
- Regression tests in `test/channel-registry-routing.test.ts`.

## Notes
- Stacked on `fix/agent-temperature-omit` (which introduces the host-owned routing / `routeSpeech` this builds on) — hence the base. Diff here is a single commit.
- Full suite green (120/120); verified live (Cairn now replies in a freshly-created channel).

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -183,6 +183,12 @@ export class AgentFramework {
   private processLoggingBroadcast: boolean;
   private activeStreams: Map<string, Promise<void>> = new Map();
   private pendingAssistantBlocks: Map<string, ContentBlock[]> = new Map();
+  /** Per-agent count of consecutive exhausted inferences (reset on any success).
+   *  Drives hard-down escalation — see noteInferenceExhausted. */
+  private consecutiveInferenceFailures: Map<string, number> = new Map();
+  /** N consecutive failed inferences ⇒ the agent is treated as hard-down and
+   *  escalated loudly to stderr. */
+  private readonly inferenceFailureEscalationThreshold = 3;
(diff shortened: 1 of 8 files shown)


antra-tess:
Superseded by #42 — the branch is contained verbatim in consolidate/adaptive-default.

PR #7, opened 2026-02-19
mcpl: implement full MCPL host support (Steps 1-8)

Tengro:
## Summary

Transforms agent-framework into a full MCPL v0.4.0-draft host, enabling MCPL servers as first-class participants alongside existing modules. This is an additive change — zero behavior change when no `mcplServers` are configured.

- **12 new files** in `src/mcpl/` implementing the complete MCPL protocol surface: server connections (JSON-RPC 2.0 over stdio), feature set management, scoped access control, context hooks, push events, server-initiated inference, channel lifecycle, and state management with checkpoint trees
- **Framework integration** in `framework.ts` (+535 lines): subsystem initialization, event wiring, tool dispatch with `mcpl:{serverId}:{toolName}` routing, beforeInference/afterInference hooks, stateful tool support
- **Agent decomposition** in `agent.ts`: composable inference steps (`compileWithInjections`, `startStreamWithInjections`) enabling context injection from both modules and MCPL servers
- **Cross-cutting cherry-picks** from `Anarchid/agent-framework@mcpl-module-proto`: `endTurn` on ToolResult (sleep-until-next-event primitive), `gatherContext()` on Module interface, reconnect with background retry

### MCPL subsystems implemented

| Subsystem | File | Spec Section |
|-----------|------|-------------|
| Types + Errors | `types.ts`, `errors.ts` | All |
| Server Connection | `server-connection.ts` | 3-4 |
| Server Registry | `server-registry.ts` | — |
| Feature Sets | `feature-set-manager.ts` | 5 |
| Scoped Access | `scope-manager.ts` | 7 |
| Context Hooks | `hook-orchestrator.ts` | 10 |
| Push Events | `push-handler.ts` | 9 |
| Server Inference | `inference-router.ts` | 11 |
| Channels | `channel-registry.ts` | 14 |
| State Management | `checkpoint-manager.ts` | 8 |

### Key design decisions

- **Null-safety pattern**: All MCPL fields are `T | null`, checked before use — zero overhead when no MCPL servers configured
- **Fail-open everywhere**: Timeouts, server errors, hook failures never block inference
- **Chronicle-backed state**: Checkpoint trees persisted to Chronicle state slots, surviving framework restarts
- **Reconnect with background retry**: Non-blocking start on initial connection failure, automatic retry on process exit

## Test plan

- [ ] `npx tsc --noEmit` passes (with targeted file check due to pre-existing membrane `composite` tsconfig issue)
[...]

PR #5, opened 2026-01-29
Fixes merged, discord app example improved

Tengro:

PR #40, opened 2026-06-09
Preserve image blocks through MCPL tool-result conversion

Anarchid:
Pairs with [antra-tess/membrane#feat/tool-result-image-content](https://github.com/antra-tess/membrane/pulls) and antra-tess/zulip_mcp#7. All three are needed for an MCPL tool to deliver an image to a vision-capable model — this PR carries the AF half.

## What changes

Two coordinated changes in `src/framework.ts`:

### 1. `dispatchMcplToolCall` — keep non-text content blocks

Before:
```ts
const textContent = result.content
  ?.filter((c) => c.type === 'text' && c.text)
  .map((c) => c.text!)
  .join('\n');
// data: textContent  ← image blocks silently dropped here
```

The async MCPL dispatch path was extracting *only* text blocks and joining them into a string before pushing the `tool-result` event. If an MCP tool returned `[{type:'text'}, {type:'image'}, …]`, the image was gone before AF's own conversion layer ever saw it.

Now: when any non-text block is present, pass the full `result.content` array through as `data`. Text-only results still collapse to a joined string for callers that expect string-ish data.

### 2. `toMembraneToolResult` — native fast-path for arrays carrying images

`toMembraneToolResult` was unconditionally `JSON.stringify`-ing every tool result, which would have collapsed any preserved image array into a JSON literal even if (1) hadn't dropped it first.

Add `tryNativeToolResultContent()` that detects an MCP content array carrying at least one image and converts to Membrane's native `ToolResultContentBlock[]`. Text-only / scalar / object results still go through `JSON.stringify`; the existing conversion tests for the JSON path continue to pass.

Bump `0.5.6` → `0.5.7-dev` so workspace-linked downstreams can pick this up via dev-link without a tagged release.

## Tests

All 20 existing framework tests pass. No new tests added in this PR — the round-trip is exercised by the end-to-end run described below.

## End-to-end validation

Tested against a live Zulip channel through the full chain (zulip-mcp → AF → membrane → Anthropic):

1. A 1.5MB Zero-K screenshot. Claude described top-down Mars terrain, gray platform with "STA: 0:0" overlay, triangular metal structures, red enemy markers — all matching the actual bytes.
2. A Zulip UI screenshot. Claude verbatim-quoted strings from the pixels including the exact filename `knowledge-requests/2026-05-12-system-check-across-fleet.md` that couldn't have been in its context.

Found *during* this validation: an earlier round of the test "passed" with Claude confidently describing a Python REPL (topic name was "round") — but pulling the image down and comparing showed it was actually a game screenshot, and the inference log showed `tool_result.content` arriving at Anthropic as a 77-char text string. That hallucination is what exposed the dispatch-path drop and motivated change (1).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@animalabs/agent-framework",
-  "version": "0.5.6",
+  "version": "0.5.7-dev",
   "description": "Multi-agent framework with pluggable modules, persistent state, and concurrent tool execution",
   "repository": {
     "type": "git",
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -1,5 +1,6 @@
 import type { Membrane, NormalizedMessage, NormalizedRequest, ContentBlock, YieldingStream } from '@animalabs/membrane';
 import { isAbortedResponse } from '@animalabs/membrane';
+import { toolResultDataToHistoryString } from './tool-result-history.js';
 
 export interface StartStreamResult {
   stream: YieldingStream;
@@ -509,13 +510,15 @@ export class Agent {
   }
 
   private buildToolResultMessages(results: CompletedToolCall[]): NormalizedMessage[] {
-    // Tool results go as a user message with tool_result blocks
+    // Tool results go as a user message with tool_result blocks. The history
+    // serializer turns MCP image blocks into `[image: type, size]` so the
+    // persisted transcript stays small and survives truncation.
     const content = results.map((r) => ({
       type: 'tool_result' as const,
       toolUseId: r.id,
       content: r.result.isError
         ? r.result.error ?? 'Unknown error'
-        : JSON.stringify(r.result.data),
+        : toolResultDataToHistoryString(r.result.data),
       isError: r.result.isError,
     }));
 
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -1,6 +1,6 @@
 import { join } from 'node:path';
 import { JsStore } from '@animalabs/chronicle';
-import type { Membrane, ContentBlock, NormalizedRequest, YieldingStream, ToolResult as MembraneToolResult } from '@animalabs/membrane';
+import type { Membrane, ContentBlock, NormalizedRequest, YieldingStream, ToolResult as MembraneToolResult, ToolResultContentBlock } from '@animalabs/membrane';
 import { MembraneError } from '@animalabs/membrane';
 import { ContextManager, PassthroughStrategy } from '@animalabs/context-manager';
 import type {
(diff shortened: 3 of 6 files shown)


Anarchid:
## 🟠 Review — works for the happy path, and that's where it gets concerning

tsc clean, 20/20 framework tests green, and the live 1.5MB-screenshot validation is genuinely solid diligence. The native-image fast-path in `toMembraneToolResult` is correct and the `mimeType → mediaType` / base64-source mapping typechecks cleanly. But "preserve image blocks through MCPL tool-result conversion" is a stronger claim than the diff earns.

### 1. The fix is half-applied — the persisted-context copy still stringifies the base64

`src/framework.ts:1396-1414` and `:1462-1465` run **in the same handler off the same `currentState.toolResults`**:

```ts
// 1396 — PERSISTED copy (stored to context manager → Chronicle)
content = JSON.stringify(tc.result.data);            // image array → base64 JSON literal
if (maxChars && content.length > maxChars) {
  content = safeSlice(content, 0, maxChars) + '...[truncated]';   // chopped mid-base64
}
agent.getContextManager().addMessage('user', toolResultContent);

// 1462 — LIVE copy sent to the model — THE PATH THIS PR FIXED
const membraneResults = currentState.toolResults.map(tc =>
  this.toMembraneToolResult(tc.id, tc.result, maxChars));
currentState.stream.provideToolResults(membraneResults);
```

`data` is now polymorphic (`string | McpToolResultContent[]`). Line 1462 learned about the array; line 1396 did not. So for an image result the **model sees the image this turn**, but the **recorded history** stores `JSON.stringify([...,{type:'image',data:'<1.5MB base64>'}])`, truncated mid-base64. The moment anything recompiles from stored messages — compression, `/restore`, a second inference in the same conversation — the image is gone and replaced by a corrupt base64 literal. That re-introduces the exact bug this PR exists to kill, deferred by one turn — in a framework built on rolling context compression.

Same blind `JSON.stringify(tc.result.data)` lives at **`framework.ts:1909-1915`** (no truncation at all) and **`agent.ts:518`** (`buildToolResultMessages`). Four serialization sites, one fixed.

**Suggested fix:** funnel all four through a single "AF ToolResult → content block(s)" conversion. For the persisted paths, prefer storing a placeholder for non-text blocks (e.g. `[image: image/png, 1.5MB]`) so history isn't poisoned, and let only the live membrane path carry real bytes.

### 2. PR body says "Two coordinated changes" — the diff has three

Commit `cc06f19` adds `logInference` to the abort/catch paths of `driveStream` (`:2119`, `:2164`, Postmortem 2026-05-28 P2 #7). The code is fine (`startTime`/`requestId`/`compiledRequest` all in scope, tsc agrees) — but it's undocumented in the description. Please split it out or describe it.

### 3. (minor) `maxChars` bypassed for image results

`framework.ts:2254-2257` returns before the truncation guard. Not chopping the *image* is right, but the **text blocks alongside it are now uncapped**. Cap the text blocks.

### 4. (minor) `isError: afResult.isError` is provably false here

`:2256`/`:2263` — line 2248 already returns on `isError`, so this is always falsy. Misleading; pass `false`.

### 5. (minor) seven `(b as any)` casts in `tryNativeToolResultContent`

Narrow once (`const blk = b as { type?: unknown; ... }`) instead of re-casting per line.

### 6. No unit tests on a pure, trivially-testable function

`tryNativeToolResultContent` is `unknown → ToolResultContentBlock[] | null` with ~5 obvious branches. The justification ("exercised by the end-to-end run") is undercut by the PR's own account of a previous E2E run that "passed" via a hallucination nobody caught until the bytes were diffed by hand — and a **two-turn** test would have caught issue #1 directly.

---

**Verdict:** smart fix, honest validation, good bones — but it preserves the image for exactly one inference and one code path. Converge all four serialization sites (or store placeholders in history), document/split the bundled logging change, and add the unit tests, and this flips to 🟢.

Anarchid:
Thanks — substantive review and the #1 finding was real. Addressed in `09d99ba` (and rebased onto current `main` to drop the accidentally-bundled `cc06f19`, so the diff is now cleanly just the image-passthrough work):

**#1 — converge all four serialization sites.** New `src/tool-result-history.ts` exports `toolResultDataToHistoryString(data, maxChars?)`. For MCP-shaped content arrays it keeps text blocks verbatim and replaces image blocks with a short `[image: mimeType, ~NMB]` placeholder; falls back to JSON for unknown shapes. The three persisted-history sites (`framework.ts:1396`, `framework.ts:1909`, `agent.ts:518`) all route through it. The live membrane path keeps its array passthrough.

**#2 — bundled `cc06f19` removed.** Rebased onto `origin/main`; the PR now contains exactly one commit (plus this follow-up).

**#3 — cap text alongside images.** `tryNativeToolResultContent` now takes `maxChars` and truncates each text block when provided, so an image next to a 50MB log dump can't blow context.

**#4 — `isError: false` explicit.** Done.

**#5 — narrow casts once.** Single `as { type?: unknown; ... }` instead of seven per-line casts.

**#6 — unit tests.** New `test/tool-result-history.test.ts` with 10 tests covering text-only / image / mixed-known / unknown / null / empty / maxChars branches. Updated the one framework test that asserted the now-explicit `isError: false`. All 30 tests pass.

E2E re-run on the same live Zulip channel after the convergence: Claude still quotes the correct strings from actual pixels, and a `path: "/api/v1/users"` attempt rejects at the zulip-mcp boundary (companion PR antra-tess/zulip_mcp#7).

Anarchid:
## 🟢 Re-review — all six addressed, flipping to approve

Pulled `09d99ba`. tsc clean, 20/20 framework tests + the new 10/10 `tool-result-history` tests green. Walked each finding:

1. **Convergence (the big one).** New `src/tool-result-history.ts → toolResultDataToHistoryString` is the single serializer, and all three persisted sites (`framework.ts:1396`, `:1909`, `agent.ts:518`) now route through it while the live membrane path keeps its array passthrough. Image blocks become a `[image: type, ~NMB]` placeholder in history instead of truncated base64 — exactly the split that keeps the model getting bytes this turn without poisoning the transcript. 👍
2. **Bundled logging change** rebased out of the PR. Clean scope now.
3. `tryNativeToolResultContent` caps each accompanying text block under `maxChars`.
4. Explicit `isError: false` on the non-error return paths.
5. Seven `(b as any)` → one narrowed `as { type?: unknown; ... }`.
6. Real unit tests across text-only / image / mixed / unknown / null / empty / maxChars.

One property worth stating out loud (not a blocker): the persisted copy is now intentionally a placeholder, so the model only sees real image bytes on the turn the result lands — including the over-budget stream-restart path, which recompiles from history and will hand the model the placeholder. That's strictly better than the previous corrupt-base64 behavior and is the right tradeoff; just flagging it as a known property in case a future "why didn't it re-see the image" question comes up.

Nice, honest turnaround — the commit message even credits the self-review and explains the failure mode. LGTM. 🟢

PR #49, opened 2026-06-14
test: add channel_publish fence coverage (home-default + foreign-reject)

### LINKED ISSUE #45: Conversation routing: channel_publish fence is untested (home-default + foreign-reject)
**Found during review of #41. Test-coverage gap on a merged feature.**

### Problem
The channel fence in `dispatchChannelToolCall` is tested only for `channel_open` and foreign `channel_close`. The action forks actually perform — **`channel_publish`** — has no coverage:
- home-defaulting (omitted `channelId` → injected `home`) — untested
- foreign-publish rejection — untested

Additionally, the existing fence test reaches through `as unknown as { dispatchChannelToolCall }` and calls the private directly, so it validates the guard in isolation, not that a fork's real tool call is *routed* through it.

### Suggested fix
Add a `channel_publish` case to `test/conversation-routing.test.ts`: assert (a) omitted `channelId` is rewritten to the home channel, and (b) a foreign `channelId` is rejected. ~3 asserts.

koteshyelamati:
Closes #45

## What

Adds two missing test cases for the `channel_publish` fence in `dispatchChannelToolCall`, as called out in issue #45:

1. **Home-defaulting** — when a fork calls `channel_publish` with no `channelId`, the fence should inject the home channel and **not** reject the call. Asserts `failures.length === 0`.
2. **Foreign-reject** — when a fork calls `channel_publish` with a foreign `channelId` (e.g. `slack:C99` when home is `slack:C12`), the fence must reject it. Asserts `failures.length === 1` and the error references the foreign channel.

Both tests follow the same pattern as the existing `'forks cannot open channels or close foreign ones'` test, calling `dispatchChannelToolCall` directly via the private cast and capturing `tool:failed` trace events.

### DIFF
--- a/test/conversation-routing.test.ts
+++ b/test/conversation-routing.test.ts
@@ -1,4 +1,289 @@
+/**
+ * Integration tests for per-channel conversation routing: incoming MCPL
+ * channel messages spawn/route to fork agents instead of the primary
+ * conversation when FrameworkConfig.conversations is set.
+ */
 
+import { describe, it, beforeEach, afterEach } from 'node:test';
+import assert from 'node:assert/strict';
+import { tmpdir } from 'node:os';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { join } from 'node:path';
+import { AgentFramework } from '../src/index.js';
+import type { ProcessEvent, ConversationRouterConfig, TraceEvent } from '../src/index.js';
+import { MockMembrane, createMockResponse } from './helpers/mock-membrane.js';
+
+function incomingEvent(overrides: {
+  channelId: string;
+  text: string;
+  mentioned?: boolean;
+  channelType?: string;
+  messageId?: string;
+}): ProcessEvent {
+  return {
+    type: 'mcpl:channel-incoming',
+    serverId: 'srv',
+    channelId: overrides.channelId,
+    messageId: overrides.messageId ?? `m-${Math.random().toString(36).slice(2)}`,
+    author: { id: 'U1', name: 'alice' },
+    content: [{ type: 'text', text: overrides.text }],
+    timestamp: new Date().toISOString(),
+    metadata: {
+      mentioned: overrides.mentioned ?? false,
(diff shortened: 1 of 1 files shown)


Anarchid (changes requested):
Thanks for picking up #45. Requesting changes — the main issue is that one of the two new tests doesn't pass.

### The home-default test fails, on this branch and after merging current main

```
not ok 9 - channel_publish with omitted channelId defaults to home channel and is not rejected
error: "Cannot read properties of null (reading 'handleChannelToolCall')"
    AgentFramework.dispatchChannelToolCall (src/framework.ts:4015:27)
```

Root cause: this suite's `makeFramework()` configures no MCPL servers, so `framework.channelRegistry` is null. The existing fence tests only exercise *rejection* paths, which return synchronously before touching the registry. Your foreign-reject test passes for the same reason. The home-default test is the first to take the *accept* path, which continues past the fence into `this.channelRegistry!.handleChannelToolCall(...)` and throws.

To fix, the accept path needs somewhere to land — either a minimal mock channel registry installed on the framework, or a real MCPL server fixture in the suite setup.

### The assertion doesn't verify what #45 asked for

`assert.equal(failures.length, 0)` runs synchronously, immediately after a call whose downstream result surfaces via an async `.then/.catch` chain — so it only proves the fence didn't synchronously reject. #45 specifically asks to assert that the omitted `channelId` is **rewritten to the home channel**; the test should capture what reaches the registry (easy once a mock registry exists) and assert `channelId === 'slack:C11'` there.

### Minor

- Please squash the two commits — the first replaces the entire file and the second restores it, and the first commit's message describes a refactor that isn't in the final diff.
- The branch is ~22 commits behind main (merges cleanly, but worth rebasing while you're in there).
- #45 also flagged that these tests reach the fence through a private cast rather than a fork's real tool-call routing. Your PR description acknowledges this; fine to leave for a follow-up, just noting it remains open.

Side note (ours, not yours): this failure exposed that `dispatchChannelToolCall` non-null-asserts `channelRegistry` on the accept path even though the fence path can run without one — we'll add a guard separately.

PR #51, opened 2026-07-02
fix: channel-scope module gatherContext injections in conversation forks

### LINKED ISSUE #43: Conversation forks: module gatherContext injections are not channel-scoped
**Found during review of #41 (per-channel conversation routing). Merged with this as a known follow-up.**

### Problem
Channel isolation is the headline guarantee of the conversation-routing feature, but it is only enforced for **hook** injections. Module `gatherContext` injections bypass `scopeInjectionsForAgent` entirely.

`src/framework.ts` (~L2079):
```ts
// NOTE: module injections are NOT channel-scoped — only the MCPL hook
// injections below pass through scopeInjectionsForAgent ...
const moduleInjections = await this.moduleRegistry.gatherContext(agent.name); // unscoped → injections
...
const hookInjections = this.scopeInjectionsForAgent(
  agent.name,
  await this.hookOrchestrator.beforeInference(hookParams),                    // scoped ✓
);
```

### Impact
Latent today (per-channel context currently arrives via the scoped `beforeInference` hook; module injections are agent-global like lessons/retrieval). The day a module emits per-channel context via `gatherContext`, every fork silently inherits other channels' context — no error, no failing test, just cross-engagement leakage.

### Suggested fix
Route **both** injection sources through `scopeInjectionsForAgent`; make opt-out the special case rather than the default path. Then delete the cautionary NOTE comment.

mvanhorn:
## Summary
Module `gatherContext` injections were applied without passing through `scopeInjectionsForAgent`, so a conversation fork could see context emitted for a different channel. Only the MCPL hook injections were being scoped.

## Changes
Route module injections through `scopeInjectionsForAgent(agentName, ...)` at the three injection sites, matching how hook injections are already scoped. Removed the stale NOTE comment that documented the unscoped behavior.

## Testing
Added `test/conversation-fork-injection-scoping.test.ts` asserting a forked agent does not receive another channel's module injections.

Fixes #43

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -836,8 +836,9 @@ export class AgentFramework {
     // Module gatherContext (fail-open, matches startAgentStream)
     try {
       const moduleInjections = await this.moduleRegistry.gatherContext(agentName);
-      if (moduleInjections.length > 0) {
-        injections = moduleInjections;
+      const scopedModuleInjections = this.scopeInjectionsForAgent(agentName, moduleInjections);
+      if (scopedModuleInjections.length > 0) {
+        injections = scopedModuleInjections;
       }
     } catch (error) {
       console.error('Module gatherContext error (preview):', error);
@@ -848,7 +849,10 @@ export class AgentFramework {
     if (this.hookOrchestrator) {
       try {
         const hookParams = this.buildBeforeInferenceParams(agent);
-        const hookInjections = await this.hookOrchestrator.beforeInference(hookParams);
+        const hookInjections = this.scopeInjectionsForAgent(
+          agentName,
+          await this.hookOrchestrator.beforeInference(hookParams),
+        );
         if (hookInjections.length > 0) {
           injections = injections ? [...injections, ...hookInjections] : hookInjections;
         }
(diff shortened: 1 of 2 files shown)


Anarchid (changes requested):
Thanks — the fix itself is correct and well-scoped. Routing module `gatherContext` injections through `scopeInjectionsForAgent` at both the real-inference and preview sites is exactly what #43 asked for, and the additional parity fix (preview wasn't scoping **hook** injections either) is a welcome catch. The five new tests cover the right edges (home filtering, non-conversation agents, preview parity, registry-absent fail-open, empty list) and all pass, as do `framework.test.ts` (21/21) and `conversation-routing.test.ts` (8/8).

One blocking issue before merge:

### `tsc --noEmit` fails — 14 new errors in the test file

`test/conversation-fork-injection-scoping.test.ts:173–231`:

```ts
const system = membrane.calls[0]!.system;        // string | undefined
assertContains(system, 'module chanA context');  // parameter expects string
```

`NormalizedRequest.system` is optional, so every `assertContains`/`assertOmits` call on it fails typecheck with `TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'`. None of these errors exist on the base commit.

Either of these resolves it:
- add `assert.ok(system)` after each extraction (narrows the type), or
- type the helpers as `(text: string | undefined, needle: string)` and treat `undefined` as a failure inside.

(Not your problem, but for the record: main currently has one unrelated pre-existing tsc error in `src/runtime/liveness-watchdog.ts` — everything above is net-new from this branch.)

Non-blocking: the test's hand-rolled `ChannelRegistry` via `as never` casts and the `FrameworkInternals` reach-through mirror what the existing suite already does, so no objection here.

Happy to merge once typecheck is clean.

Anarchid:
Excellent - that just leaves the merge conflict :)

mvanhorn:
Conflict resolved and rebased onto main. Two things came up in the rebase: main's newer per-module timeout comment is kept (with the note updated since module injections are now scoped), and the test helpers now accept the optional system field from the newer request type. Build is clean and all 5 scoping tests pass.

PR #53, opened 2026-07-06
feat(agent): configurable prompt-cache TTL (cacheTtl)

LariTesserae:
## Problem

Membrane supports `request.cacheTtl` (`'5m' | '1h'`) end-to-end — it lands in Anthropic `cache_control.ttl` — but agent-framework never sets the field, so every deployment runs on the provider's 5-minute default.

For persistent agents whose reply cadence is slower than 5 minutes, the cache expires between turns and the **full context is re-written to cache on nearly every call**. Cache writes carry a premium (1.25× base input at 5m TTL) while reads cost 0.1×, so the write premium is paid over and over.

**Measured** on a day-one production resident (~130k live context, replies minutes apart): 26 calls → 56k uncached input / 5.48M cache reads / **3.49M cache writes** / 36k output. Cache writes were **~85% of total spend** (~$43 of ~$51).

## Change

- `AgentConfig.cacheTtl?: '5m' | '1h'` — optional, documented with the economics tradeoff (1h writes cost 2× vs 1.25×, but once per idle-hour instead of on every post-gap reply; keep 5m/unset for sub-5-minute loops).
- `Agent` stores it and `buildRequest` forwards it next to `promptCaching: true`. Unset ⇒ behavior unchanged.

Verified in production via a dist-patch of the same two lines: with `'1h'`, an equivalent conversation pattern drops to roughly a quarter of the cost; a 10-minute heartbeat becomes ~one write + N reads per hour.

## Notes

- Tests: 218 pass / 3 fail — identical counts on clean HEAD (pre-existing MCPL reconnect-lifecycle failures), change is test-neutral.
- Host plumbing (recipe field → AgentConfig) is a two-line follow-up in connectome-host; happy to open it as a companion PR.

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -46,6 +46,8 @@ export class Agent {
   readonly thinking: AgentConfig['thinking'];
   /** Refusal auto-rewind policy (see AgentConfig.refusalHandling). */
   readonly refusalHandling: AgentConfig['refusalHandling'];
+  /** Prompt-cache TTL forwarded to the provider (see AgentConfig.cacheTtl). */
+  readonly cacheTtl: AgentConfig['cacheTtl'];
 
   private _state: AgentState = { status: 'idle' };
   private _inferenceStartedAt = 0;
@@ -73,6 +75,7 @@ export class Agent {
     this.temperature = config.temperature;
     this.thinking = config.thinking;
     this.refusalHandling = config.refusalHandling;
+    this.cacheTtl = config.cacheTtl;
     this.maxStreamTokens = config.maxStreamTokens ?? 150_000;
     this.contextBudgetTokens = config.contextBudgetTokens;
     this.contextManager = contextManager;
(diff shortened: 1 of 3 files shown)


Anarchid:
Reviewed and verified locally: Membrane `^0.5.64` floor does include `cacheTtl` (checked the 0.5.64 tarball), the unset⇒unchanged claim holds in `membrane.ts`, and the suite is test-neutral against clean main (identical failures both sides — the specific failing tests differ by environment, but neutrality reproduces).

Pushed two small follow-ups to the branch before merging:
- `test/agent-cache-ttl.test.ts` — pins that `buildActivationRequest` forwards `cacheTtl` when set and omits it when unset, since the failure mode of this plumbing regressing is silent cost inflation.
- Trimmed the dated production measurement out of the `AgentConfig` doc comment (the decision rule stays; the war story lives here in the PR).

Thanks — this one was a pleasure to review. Happy to take the connectome-host companion PR as well.

Issue #47, opened 2026-06-14
Conversation forks: bot self-messages can refresh idle TTL and prevent expiry


Anarchid:
**Found during review of #41. Follow-up; adapter-dependent.**

### Problem
`route()` refreshes `lastActivity` on every incoming `channel-incoming` message (`src/mcpl/conversation-router.ts:162`). If a platform adapter echoes the bot's own published messages back as incoming — and self-filtering is a per-adapter concern, not guaranteed — then a fork that posts to its channel resets its own idle TTL. An engagement that talks to itself never expires.

### Suggested fix
Don't refresh TTL on messages authored by the bot. Either filter self-authored messages before `route()`, or pass an `isSelf`/author signal into `IncomingMessageFacts` and skip the `lastActivity` bump.

PR #54, opened 2026-07-07
fix(mcpl): reconnect re-registration, tools/call timeout, poison-history quarantine

Tengro:
Part of the July 2026 fragility-audit remediation (see UNTESTED-ROUTES-TEST-PLAN.md §6.2, §6.3, §6.5, §6.6, §3.5).

## Fixes (all 5)
- **6.3 [H] MCPL reconnect deregistration + checkpoint destruction** — factored feature-set/scope/checkpoint init out of `connectMcplServerInternal` into idempotent `registerMcplServerFeatures()`, re-run on `'reconnect'`. `'close'` skips `checkpointManager.removeServer` when `connection.willReconnect`; `disconnectMcplServer` destroys state explicitly. Fixes the silent post-reconnect deafness (pushes/inference-requests rejected) and the transient-blip checkpoint deletion. (Found + handled a latent hole: an already-transiently-closed connection emits no second `close`.)
- **6.2 [H] No tools/call timeout** — per-request timer in `server-connection.ts` (default 60s, `McplServerConfig.requestTimeoutMs`, 0 disables, `unref`'d); late responses ignored; rejection maps to an `isError` tool_result so the turn completes instead of hanging.
- **6.5 [H] Log-only "circuit breaker"** — at streak≥3 with `retryable === false`, `quarantinePoisonedHistory()` sheds the newest complete exchange via the existing `/unstick` primitives (no orphaned tool_use/thinking) and queues a retry; capped by `refusalHandling.maxRewinds`. Retryable/unknown failures never shed history.
- **6.6 [M-H]** — `JSON.stringify(afResult.data) ?? ''` in `toMembraneToolResult`.
- **3.5 [M]** — new `Module.contextTimeoutMs` per-module gatherContext budget (the contract FKM's retrieval module consumes); registry default 5s → 15s; fail-open kept, race timer no longer leaks.

## Tests / build
`tsc --noEmit` + build clean; **261/261** (21 new across 5 files, incl. a real stdio child that swallows tools/call, and reconnect re-registration through the real wiring). Version 0.6.1 → 0.6.2 (not published).

## Not verified locally (documented)
Full-stack kill-9 reconnect drill (covered at the handler boundary instead) and live behavior against a real membrane-classified 400 — both belong to the post-deploy checklist. Baseline `main` had a pre-existing fixture-copy test failure (tsc doesn't copy the `.mjs` fixture into `dist`); fixed here by resolving the fixture from source.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@animalabs/agent-framework",
-  "version": "0.6.1",
+  "version": "0.6.2",
   "description": "Multi-agent framework with pluggable modules, persistent state, and concurrent tool execution",
   "repository": {
     "type": "git",
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -219,6 +219,13 @@ interface RewindRecord {
   discordRef?: { channelId: string; messageId: string };
 }
 
+/** Cap an API error message for inline use in a rewind marker: keep enough to
+ *  identify the failure class without pasting a wall of provider JSON into the
+ *  agent's context. */
+function truncateReason(reason: string, max = 160): string {
+  return reason.length <= max ? reason : reason.slice(0, max) + '…';
+}
+
 export class AgentFramework {
   private store: JsStore;
   private ownsStore: boolean;
@@ -258,6 +265,11 @@ export class AgentFramework {
    *  chain (reset when a turn completes without a refusal). Bounds the auto
    *  rewind loop — see refusalHandling + rewindTriggeringTurn. */
   private refusalRewinds: Map<string, number> = new Map();
+  /** Per-agent count of poison-history rewinds performed by the hard-down
+   *  breaker (noteInferenceExhausted). Reset on any successful inference.
+   *  Bounds the automatic quarantine loop the same way refusalRewinds bounds
+   *  the refusal loop, so the breaker can never shed the whole history. */
+  private exhaustionRewinds: Map<string, number> = new Map();
   /** Per-agent current rewind episode: the single consolidated marker's id and
    *  how many turns have been shed so far. One marker per episode, updated in
    *  place (see updateRewindMarker); cleared when the episode ends. */
@@ -1441,15 +1453,24 @@ export class AgentFramework {
    * sits at the tail giving the model something actionable to answer once the
    * refusal clears. Returns the running count.
    */
-  private updateRewindMarker(agent: Agent, category: string): number {
+  private updateRewindMarker(
+    agent: Agent,
+    category: string,
+    cause: 'refusal' | 'inference-failure' = 'refusal',
+  ): number {
     const cm = agent.getContextManager();
     const ep = this.rewindEpisode.get(agent.name);
     const count = (ep?.count ?? 0) + 1;
+    const why = cause === 'refusal'
+      ? `the model refused on them (content filter: ${category}). Their content is ` +
+        `not reproduced here (so this note can't re-trigger the filter); the ` +
+        `originals remain in the raw record`
+      : `the model API kept rejecting the conversation on them (${category}). ` +
+        `Their content is not reproduced here (so this note can't re-trigger the ` +
+        `rejection); the originals remain in the raw record`;
     const text =
       `[refusal-rewind] The ${count} most recent turn(s) were set aside because ` +
-      `the model refused on them (content filter: ${category}). Their content is ` +
-      `not reproduced here (so this note can't re-trigger the filter); the ` +
-      `originals remain in the raw record. You are clear to continue — if you ` +
+      `${why}. You are clear to continue — if you ` +
       `were mid-task, take a different approach; otherwise carry on with whatever ` +
       `is now in front of you, or briefly acknowledge the gap and ask what's next.`;
     const blocks: ContentBlock[] = [{ type: 'text', text }];
@@ -1458,7 +1479,7 @@ export class AgentFramework {
       this.rewindEpisode.set(agent.name, { markerId: ep.markerId, count, category });
     } else {
       const id = cm.addMessage('user', blocks, {
-        system: true, kind: 'refusal-rewind', category, count,
+        system: true, kind: 'refusal-rewind', category, count, cause,
       });
       this.rewindEpisode.set(agent.name, { markerId: id, count, category });
     }
@@ -3051,7 +3072,8 @@ export class AgentFramework {
       // Both produce ContextInjection[] that get merged before inference.
       let injections: ContextInjection[] | undefined;
 
-      // Module gatherContext (fail-open, 5s timeout per module)
+      // Module gatherContext (fail-open, per-module timeout — the module's
+      // contextTimeoutMs, else the registry default)
       // NOTE: module injections are NOT channel-scoped — only the MCPL hook
       // injections below pass through scopeInjectionsForAgent (channel
       // context arrives via beforeInference hooks by adapter convention). A
@@ -3105,6 +3127,9 @@ export class AgentFramework {
           type: 'inference:exhausted',
           agentName: agent.name,
           error: err.message,
+          // Drives the poison-history breaker: only a known NON-retryable
+          // failure (e.g. a 400 on a corrupt history) triggers auto-rewind.
+          retryable: err instanceof MembraneError ? err.retryable : undefined,
         });
         this.eventGate?.onInferenceEnded(agent.name);
         if (action.emit) {
@@ -3612,6 +3637,9 @@ export class AgentFramework {
                 type: 'inference:exhausted',
                 agentName: agent.name,
                 error: err.message,
+                // Drives the poison-history breaker: only a known NON-retryable
+                // failure (e.g. a 400 on a corrupt history) triggers auto-rewind.
+                retryable: err instanceof MembraneError ? err.retryable : undefined,
               });
               this.eventGate?.onInferenceEnded(agent.name);
               if (action.emit) {
@@ -3759,7 +3787,10 @@ export class AgentFramework {
     if (blocks) {
       return { toolUseId: callId, content: blocks, isError: false };
     }
-    let content = JSON.stringify(afResult.data);
+    // JSON.stringify returns the VALUE undefined (not a string) for undefined
+    // input — a module tool returning `{ success: true }` with no data would
+    // otherwise make `content.length` below throw a TypeError mid-turn.
+    let content = JSON.stringify(afResult.data) ?? '';
     if (maxChars && content.length > maxChars) {
       content = safeSlice(content, 0, maxChars)
         + '\n\n[truncated — original was ' + content.length + ' chars]';
@@ -4089,13 +4120,18 @@ export class AgentFramework {
       this.noteInferenceExhausted(
         (event.agentName as string) ?? 'unknown',
         (event.error as string) ?? 'unknown error',
+        event.retryable as boolean | undefined,
       );
     } else if (event.type === 'inference:completed') {
       // A successful response — even mid-turn between tool calls — proves the
-      // agent isn't hard-down; clear its consecutive-failure streak.
+      // agent isn't hard-down; clear its consecutive-failure streak (and the
+      // poison-history breaker's rewind budget).
       const name = event.agentName as string | undefined;
       if (name && this.consecutiveInferenceFailures.get(name)) {
         this.consecutiveInferenceFailures.set(name, 0);
+      }
+      if (name && this.exhaustionRewinds.get(name)) {
+        this.exhaustionRewinds.set(name, 0);
       }
     }
 
@@ -4127,8 +4163,17 @@ export class AgentFramework {
    *      inference, so this never causes a retry/wake loop.
    *   3. escalation — after N consecutive failures the agent is hard-down;
    *      log that loudly (a repeated identical failure is the textbook signal).
+   *   4. breaker — when the failures are known NON-retryable (a 400-class
+   *      rejection of the history itself, e.g. corrupted tool_use/tool_result
+   *      pairing or an oversized attachment), retrying the same context can
+   *      never succeed: every new push event wakes the agent onto the same
+   *      poisoned history forever. At the hard-down threshold, automatically
+   *      quarantine: shed the newest complete exchange (the same forced-rewind
+   *      primitive `/unstick` uses — shedNewestTurn never orphans a
+   *      tool_use/thinking block) and retry, bounded by the same rewind cap so
+   *      the breaker can never eat the whole history.
    */
-  private noteInferenceExhausted(agentName: string, reason: string): void {
+  private noteInferenceExhausted(agentName: string, reason: string, retryable?: boolean): void {
     const streak = (this.consecutiveInferenceFailures.get(agentName) ?? 0) + 1;
     this.consecutiveInferenceFailures.set(agentName, streak);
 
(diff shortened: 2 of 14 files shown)


Tengro:
## 🔥 Roast Review (automated, Claude)

### Overall severity: 🟠 CONCERNING

The three fixes are real, the tests are unusually honest (an actual child process that swallows `tools/call` — thank you for not mocking that), tsc is clean, and every PR-touched test passes. And yet. The poison-history breaker's trigger predicate is a category error that will shed healthy history the first time an API key expires, and the checkpoint-preservation fix — the headline of §6.3 — still destroys every checkpoint tree on a *graceful* shutdown. You fixed the crash case and left the `stop()` case, which means `kill -9` now preserves data that a clean restart erases. I need you to sit with that sentence for a moment.

---

### 🟠 CONCERNING

#### 1. `retryable === false` is not a synonym for "poisoned history" — the breaker sheds good history on auth failures, aborts, and context overflow

`src/framework.ts:4212-4216`:

```ts
// (4) Poison-history breaker: only for known NON-retryable failures —
// a transient outage (overloaded/5xx) must never cost history.
if (retryable === false && agent) {
  this.quarantinePoisonedHistory(agent, reason);
}
```

The PR body says "a 400 on a corrupt history, e.g. corrupted tool_use/tool_result pairing." The code says "anything membrane classifies as non-retryable." Those are not the same set. From `membrane/src/types/errors.ts`, `retryable: false` covers:

- `auth` (401) — errors.ts:166-175. **An expired/rotated API key produces 3 consecutive non-retryable failures, and the breaker starts eating the agent's newest exchanges** — up to `maxRewinds` of them — while stamping a marker telling the agent "the model API kept rejecting the conversation on them." No it didn't. The API rejected your *credentials*. History was fine; now it's quarantined and the context is polluted with a false confession.
- `abort` (errors.ts:208-216) — a deliberately cancelled request is not history poison.
- `context_length` (errors.ts:144-153) — here shedding is actively the *wrong direction*: compression should handle overflow, and if you must shed, the newest exchange is the most valuable one and the least likely to fix the overflow.
- `safety`, `unsupported` — also swept in.

Yes, the cap bounds the damage at 3 and the originals survive in the chronicle raw record. That makes this recoverable, not correct.

**Fix**: membrane already gives you the discrimination you need — `MembraneError.type`. Forward it alongside (or instead of) `retryable` at the two emit sites (framework.ts:3132, :3640) and gate the breaker on `type === 'invalid_request'` (arguably plus `context_length` with an *oldest*-first strategy, but that's a different PR). One line at each emit site, one changed condition at :4214. Your own test's `REASON` string is literally a 400 `invalid_request` — the tests encode the intent the code doesn't enforce.

#### 2. Graceful shutdown still permanently destroys every checkpoint tree — the `willReconnect` discriminator conflates "won't reconnect" with "permanently removed"

The chain, all verified in-source:

- `AgentFramework.stop()` → `mcplServerRegistry.closeAll()` (framework.ts:622)
- → `connection.close()`, which sets `reconnectEnabled = false` **before** emitting `'close'` (server-connection.ts:473, :498)
- → your new close handler, framework.ts:4795-4798:

```ts
this.featureSetManager?.removeServer(connection.id);
if (!connection.willReconnect) {
  this.checkpointManager?.removeServer(connection.id);
}
```

- → `CheckpointManager.removeServer` deletes the trees **and calls `persistTree()`** (checkpoint-manager.ts:449-451), writing the deletion to Chronicle before the store closes.

So the comment you added — "checkpoint trees are durable state … a single crash/blip [must not] permanently destroy every checkpoint" — is aspirational. A SIGKILL'd host preserves checkpoints (no close handler runs with `willReconnect` false); a polite `stop()` erases them, durably. Host restart is the *most common* lifecycle event there is, and it's precisely the one where "durable state persisted to Chronicle" is supposed to pay off (`loadFromStore()` exists for exactly this — it just finds the tombstone you wrote on the way out).

This destruction was pre-existing (the old handler removed unconditionally), but this PR is the designated fix for this exact class, rewrote this exact line, and added the comment claiming durability. Half-fixing it with a confident comment is worse than not touching it.

**Fix**: the close handler should *never* destroy checkpoints. Permanent removal already has an explicit owner — `disconnectMcplServer` (framework.ts:4570-4571), which you correctly made do it directly. Drop the `if (!connection.willReconnect) removeServer` from the close handler entirely, and your fourth re-registration test still passes because it goes through `disconnectMcplServer`.

---

### 🟡 MEH

#### 3. `retryable` never made it into the `TraceEvent` type — it's smuggled through the index signature and read back with a cast

`src/types/trace.ts:59-63` is unchanged:

```ts
| (TraceEventBase & {
    type: 'inference:exhausted';
    agentName: string;
    error: string;
  })
```

while framework.ts:4123 does `event.retryable as boolean | undefined`. Typed trace consumers can't see the field; the compiler can't see the contract; a typo (`retriable`) at either end would pass tsc and silently disable the breaker. Add `retryable?: boolean` to the union member — it's four lines.

Related, and worth saying out loud: this field now *drives work* — history mutation and a queued inference request — from inside `emitTrace`. The project's own line is ProcessEvents = work, TraceEvents = observability. The interception predates this PR, granted, but the PR promotes the trace channel from "logs health" to "mutates the chronicle," which deepens the conflation. Both emit sites could call `noteInferenceExhausted(...)` directly and let `emitTrace` go back to being a reporter.

#### 4. A timed-out `tools/call` is "abandoned," not cancelled — late results of *stateful* tools are silently discarded, checkpoints included

`server-connection.ts:708-712`:

```ts
const pending = this.pendingRequests.get(response.id);
if (!pending) {
  return; // Orphaned response — ignore
}
```

Fine for stateless tools. But the framework's stateful dispatch (framework.ts:4959-4977) records `result.state` checkpoints from the response — and an orphaned late response carrying a checkpoint is dropped on the floor. The server advanced its checkpoint tree; the host didn't; every subsequent call now sends stale `state`/`checkpoint` params (framework.ts:4946-4957). Silent divergence, no trace event, nothing to grep for.

Also: the timeout error the *model* sees says "The request was abandoned; the connection remains open" — nothing about the tool possibly still completing server-side. The model's rational next move is to retry, which for `send_message`-class tools means duplicates. Cheap fixes: (a) emit an `mcpl:orphaned-response` trace (minimum) or route late stateful results into `recordCheckpoint`; (b) append "the tool may still have completed server-side — verify before retrying" to the timeout message at server-connection.ts:599-603. The words the model reads are load-bearing; you know this.

#### 5. `contextTimeoutMs` is unclamped — any module can now legally hold every inference turn hostage for a self-declared budget

`src/module-registry.ts:319`:

```ts
const budgetMs = module.contextTimeoutMs ?? timeoutMs;
```

No ceiling. A module declaring `contextTimeoutMs: 600_000` that wedges makes *every turn of every agent* wait ten minutes, and both call sites (framework.ts:808, :3083) use the defaults. The fail-open principle you cite in the docstring is "timeouts never block inference" — this feature converts it to "timeouts block inference for exactly as long as the module asks." The FKM retrieval use case is legitimate; the missing `Math.min(budgetMs, SOME_CEILING)` is not. Also noted without further comment: the shared default silently tripled (5s → 15s) for every existing deployment, so a wedged module now costs 3x per turn out of the box.

#### 6. The `Object.create(AgentFramework.prototype)` test harnesses are runtime-fragile, and this PR is its own evidence

You had to patch `test/inference-failure-observability.test.ts` to hand-add `exhaustionRewinds`, `rewindEpisode`, and `pendingRequests` because the new code paths touch private fields the old harness didn't stub. That's the failure mode on a loop: every future private field consulted by `noteInferenceExhausted` breaks N harnesses at runtime, not compile time, with an `undefined.get is not a function` three stack frames from the cause. Three test files now build framework fragments by hand (`inference-exhaustion-rewind`, `inference-failure-observability`, `mcpl-reconnect-reregistration`). Extract one shared `makeFrameworkHarness()` in `test/helpers/` before there are five.

---

### Things checked that survived scrutiny (credit where due)

- **Reconnect re-registration is idempotent all the way down**: `FeatureSetManager.initializeServer` replaces via `Map.set` (feature-set-manager.ts:185), `ScopeManager.configureAll` overwrites (scope-manager.ts:124-128), `CheckpointManager.registerFeatureSet` early-returns on an existing tree (checkpoint-manager.ts:217). Rapid flap double-registration: not a thing. Backoff caps (exponent ≤ 30, `reconnectMaxIntervalMs`, ±25% jitter) and resets on success (server-connection.ts:556).
- **Timeout timer lifecycle is correct on all three settle paths** — response (:715), explicit close (:486), transport close (:768) all clear the timer; late responses hit the orphan guard; `unref()` so the watchdog can't pin the process. In-flight `tools/call` at transient close rejects with a descriptive error that both dispatch paths (framework.ts:3769-3775, :5009-5020) convert to an `isError` tool_result. Fail-open holds.
- **The quarantine preserves tool_use/tool_result pairing** for the actual message shape this framework produces: all of a turn's tool results land in *one* user message (framework.ts:2376-2384), so `shedNewestTurn`'s "tool_result + the assistant message before it" invariant is sound, and an orphaned trailing `tool_use` is itself shed cleanly as `kind: 'other'`.
- **Breaker discipline**: never fires on `retryable: true`/`undefined`, cap respected, budget resets on `inference:completed`, degrades to logging when only system markers remain, and the retry is picked up because the event loop polls `pendingRequests` every iteration and `agent.reset()` runs before the exhausted emit. All verified against the tests, which genuinely test it.
- All three fixes have regression tests, including a real stdio child for the hang. This is the part of the PR I have no complaints about, which pains me.

---

### Tooling results

`npx tsc --noEmit` — **clean** (exit 0).

`node --import tsx --test test/*.test.ts` (full suite, PR head):

```
tests 242 / pass 236 / fail 6 / cancelled 0 / skipped 0
```

The 6 failures are all in files this PR does not touch — `liveness-watchdog.test.ts` (3: expected SIGKILL/SIGABRT, got null; healthy process exited 1) and `event-gate-script.test.ts` (3: gate.js programmable gate + script fallthrough). I checked out `main` in a scratch worktree and ran the same files: **identical 6 failures on baseline** (1/7 pass), so they are pre-existing and environment-shaped (sandboxed signal delivery / script spawn), not regressions from this branch.

All 8 PR-relevant test files run green in isolation:

```
tests 48 / pass 48 / fail 0
```

(The PR body claims 261/261; I count 242 here — presumably a node version counting subtests differently. The delta is not hiding a failure; I checked.)

---

### Verdict

The engineering instincts are right: idempotent re-registration through the real wiring, a watchdog that cleans up after itself on every path, a breaker with a cap and a budget reset, tests that spawn actual misbehaving processes. This is the good kind of fragility work. But finding #1 means the breaker will fire on the first expired API key in production and quarantine three perfectly good exchanges while blaming the history, and finding #2 means the checkpoint durability this PR advertises evaporates on every clean restart. Both are small, targeted fixes — gate on `MembraneError.type`, delete one conditional from the close handler. Do those two and this is a 🟢 I'd be quietly pleased about. Ship it before then and §6.3 will be back in the next audit with a new section number.

Tengro:
Addressed the review (commit 3c35f7d).

**CONCERNING (both fixed):**
1. **Breaker no longer sheds good history on non-`invalid_request` failures.** The quarantine now gates on membrane's error *type* (`errorType === 'invalid_request'`), not `retryable === false`. `MembraneError.type` is forwarded alongside `retryable` at both `inference:exhausted` emit sites. auth/abort/context_length/safety/unsupported are all non-retryable but no longer touch history. Added a test asserting each of those types sheds nothing and stamps no false rejection marker; the existing tests were realigned to pass `'invalid_request'` (matching their REASON string, a 400).
2. **Close handler never destroys checkpoints.** Dropped the `if (!connection.willReconnect) removeServer` from the `'close'` handler entirely — `stop()` clears reconnect before emitting close, so gating on it erased every checkpoint tree on a clean restart while a SIGKILL preserved them. Permanent removal stays owned by `disconnectMcplServer`. The reregistration test now asserts a `willReconnect=false` close *preserves* the tree; the disconnect test still covers destruction.

**MEH:**
3. Typed `retryable?`/`errorType?` added to the `inference:exhausted` TraceEvent member (no more index-signature smuggling).
4. Late *stateful* orphaned `tools/call` responses now emit `mcpl:orphaned-response`; timeout message warns the model the tool may have completed server-side (verify before retrying).
5. `contextTimeoutMs` clamped to a 30s ceiling.

Deferred: finding 6 (shared `makeFrameworkHarness()`) — skipped as optional/risky per the review's own caveat.

tsc clean; all PR-relevant tests green (29/29 in the touched files). The 6 full-suite failures remain the pre-existing liveness-watchdog/event-gate-script environment-shaped ones — untouched, no new failures.

Tengro:
## 🔥 Roast Re-Review (automated, Claude)

Re-review of the updated head `3c35f7db` ("fix(mcpl): gate poison-breaker on error TYPE; stop close handler destroying checkpoints"), pushed after this morning's review of `1c5d60ed`. I came back expecting the usual "addressed feedback" commit that addresses the commit message and nothing else. That is not what happened, and I am mildly disoriented.

### Overall severity: 🟡 MEH — both mandatory findings are actually fixed, with regression tests. What remains is comment rot and one blind spot.

---

### Prior findings — verdicts

| # | Prior finding | Verdict |
|---|---|---|
| 1 | 🟠 Breaker gated on `retryable === false`, sweeping in auth/abort/context_length/safety/unsupported — an expired API key shed healthy history | ✅ **FIXED** |
| 2 | 🟠 Graceful `stop()` destroyed checkpoint trees (`reconnectEnabled=false` before `'close'` → `willReconnect` false → persisted deletion); `kill -9` preserved them | ✅ **FIXED** |
| 3 | 🟡 `retryable` smuggled into `inference:exhausted` via index-signature cast instead of the type | ✅ **FIXED** |
| 4 | 🟡 Late tools/call responses silently dropped state; timeout message didn't warn the model the tool may still complete | ✅ **FIXED** |
| 5 | 🟡 `contextTimeoutMs` unclamped — a module could self-declare an arbitrary budget | ✅ **FIXED** (residual note below) |
| 6 | 🟡 `Object.create(AgentFramework.prototype)` test-harness fragility | ❌ **UNFIXED** |

**1 — breaker gate.** `src/framework.ts:4234` now reads `if (errorType === 'invalid_request' && agent)`, and `'invalid_request'` is a real member of membrane's `MembraneErrorType` union (`node_modules/@animalabs/membrane/dist/types/errors.d.ts:9`), not a string someone remembered wrong. Both retry-exhaustion emit sites (`framework.ts:3131`, `:3649`) forward `errorType: err instanceof MembraneError ? err.type : undefined`. The new test `non-retryable but NON-poison failures (auth/context_length/abort) NEVER shed history` loops all five non-poison types past the hard-down threshold and asserts zero sheds, zero rewind-retries, zero false markers. This is the exact regression test I would have demanded. Passing.

**2 — checkpoint destruction.** The `'close'` handler (`framework.ts:4836`) now removes only feature-set state; the `if (!connection.willReconnect) this.checkpointManager?.removeServer(...)` block is gone, and the replacement comment correctly explains *why* `willReconnect` was an unusable signal. `disconnectMcplServer` (`framework.ts:4592`) is the sole owner of permanent removal. The old test asserting deletion-on-permanent-close was inverted into `the close handler NEVER destroys checkpoints — even with willReconnect=false`, which asserts the persisted tree survives for `loadFromStore()`. This also defuses the amplification I flagged: connectome-host PR #47's SIGTERM→`framework.stop()` conversion no longer erases every checkpoint tree on a polite restart. Fixed properly, not papered over.

**3.** `src/types/trace.ts:63-70` — `retryable?: boolean` and `errorType?: string` are now declared on the `inference:exhausted` member with doc comments. The casts remaining at `framework.ts:4130-4131` are inside `emitTrace`, whose parameter is deliberately loose-typed; acceptable.

**4.** `server-connection.ts:711-733` — a late response carrying `state`/`checkpoint` now emits `'orphaned-response'`, forwarded as an `mcpl:orphaned-response` TraceEvent (`framework.ts:4802`, typed at `trace.ts:315-323`). The timeout message now says "the tool may still have completed server-side … verify state before retrying, as a blind retry of a stateful/side-effecting tool may duplicate it." Two new tests cover the wording and the stateful-orphan event, including that the connection still works afterwards.

**5.** `module-registry.ts:36` adds `MAX_CONTEXT_TIMEOUT_MS = 30_000` and clamps at `:333`. The new test uses `mock.timers` so it doesn't actually sit through 30 seconds — thank you for not making CI pay for the clamp. *Residual:* the shared default is still 15s vs main's 5s (a silent 3× on every module without a declared budget), now at least documented in the doc comment. I'll allow it, grudgingly.

**6.** `test/inference-exhaustion-rewind.test.ts:45` and `test/mcpl-reconnect-reregistration.test.ts:68` still conjure the framework via `Object.create(AgentFramework.prototype) as any` and hand-assemble private fields. Every private-field rename in `framework.ts` will now break these tests silently-at-runtime instead of loudly-at-compile. It was a MEH and remains a MEH — but it's now a MEH with six tests standing on it.

---

### New findings (in the fix commit)

🟡 **`src/framework.ts:3700-3704` — the catch-path `inference:exhausted` emit forwards neither `retryable` nor `errorType`.**
```ts
      this.emitTrace({
        type: 'inference:exhausted',
        agentName: agent.name,
        error: err.message,
      });
```
There are four emit sites, not the "both" the commit message counts: 3127 and 3641 (instrumented), 3666 (`aborted` — correctly bare, an abort must never trip the breaker), and this one, the catch for "stream itself threw (unexpected)". If a `MembraneError` ever propagates as a throw instead of an `'error'` stream event — and this codebase has already met mid-stream SSE errors that surface through unexpected paths — a genuine `invalid_request` through here never trips the breaker. The failure direction is safe (no shedding, ever), but it's an inconsistency wearing a trench coat. Two lines: same `err instanceof MembraneError` forwarding as the other sites.

🟡 **`src/framework.ts:4587-4590` — the fix left a stale comment contradicting itself.**
```ts
    // Permanent removal: also destroy feature-set and checkpoint state
    // explicitly. The 'close' handler usually does this, but a connection that
    // already transiently closed (reconnect pending) emits no second 'close'
    // from close(), and the transient path deliberately preserves checkpoints.
```
The `'close'` handler no longer "usually does this" for checkpoints — this very commit removed that, and the close-handler comment forty lines down now declares `disconnectMcplServer` the sole owner. Two comments in the same file describing two different ownership models is how the next contributor reintroduces finding #2. Update the paragraph.

🟡 **`restartMcplServer` still destroys checkpoint trees — pre-existing (from `bbddf3c` on main), but this PR sharpened the inconsistency.** The lifecycle matrix is now: server crash → checkpoints preserved; `kill -9` the host → preserved; graceful `stop()` → preserved (fixed here); `restartMcplServer(id)` — the thing an operator calls to *unwedge a hung server* — → `disconnectMcplServer` → `checkpointManager.removeServer()` → permanent, Chronicle-persisted deletion (`framework.ts:4616`, `:4592`). The one recovery action a human deliberately takes is the only one that erases the server's durable memory. Not this PR's bug to fix, but it deserves a follow-up issue before someone learns it in production.

Nano-nits, no action demanded: `void retryable;` at `framework.ts:4233` is a strange way to spell `_retryable`; `errorType?: string` in trace.ts could be membrane's `MembraneErrorType` if you're willing to take the type-only import (the decoupling argument is defensible).

---

### Tooling results (head `3c35f7db`, run in place, clean tree)

```
$ npx tsc --noEmit
(clean, exit 0)

$ node --import tsx --test test/inference-exhaustion-rewind.test.ts \
    test/mcpl-reconnect-reregistration.test.ts \
    test/mcpl-request-timeout.test.ts test/gather-context-timeout.test.ts
ℹ tests 21  ℹ pass 21  ℹ fail 0

$ node --import tsx --test test/framework.test.ts
ℹ tests 21  ℹ pass 21  ℹ fail 0

$ node --import tsx --test test/*.test.ts
ℹ tests 246  ℹ pass 240  ℹ fail 6
✖ liveness-watchdog ×3, event-gate-script ×3
```
The 6 failures are exactly the pre-existing baseline set (liveness-watchdog ×3, event-gate-script ×3) — environment-shaped, present on main, not this PR's doing.

---

### Verdict

Both 🟠 findings were fixed at the root rather than at the symptom — the breaker now keys on the one error type that actually means "your history is poisoned," and checkpoint destruction has a single, explicit owner — and each fix arrived with the regression test that would have caught the original bug. The commit message even correctly refuses credit for the pre-existing test failures. What remains is one uninstrumented emit site (fail-safe, two-line fix), one stale comment actively contradicting the new design, the `restartMcplServer` checkpoint wipe as a follow-up, and the `Object.create` harness I will apparently be complaining about at my retirement party. None of it blocks merge. Ship it — and file the restart follow-up before you forget.

PR #55, opened 2026-07-08
refactor: settle runEphemeralToCompletion from the stream driver, not trace events

### LINKED ISSUE #33: Trace events have become load-bearing for control flow in runEphemeralToCompletion
## Background

`framework.runEphemeralToCompletion` subscribes to `inference:completed`, `inference:turn_ended`, and `inference:exhausted` via `onTrace(...)` and drives control flow from them: it resolves/rejects the caller's promise, deregisters the ephemeral agent from `this.agents`, and decides when the subagent's task is done.

This silently violates the trace-vs-process-event contract. Traces are observability — UI, logging, postmortems — and explicitly **not** the channel for framework business logic. Once the framework consumes its own traces for state transitions, the trace stream's *timing and the state observable at emission time* become part of an undocumented API surface.

## How it bit us

[PR #32](https://github.com/anima-research/agent-framework/pull/32) fixes a zombie-subagent bug surfaced in a production postmortem: a clerk agent's forked subagent timed out on the parent side, stayed registered in `this.agents`, kept receiving `mcpl:channel-incoming` fan-out for ~30 hours, and eventually duplicated a Zulip reply to an end user.

The proximate cause was an earlier refactor that moved `emitTrace({type: 'inference:completed'})` ahead of `agent.reset()` in `driveStream`. Pure observability reorder, no behavior change — except `runEphemeralToCompletion`'s trace listener does:

```ts
case 'inference:completed': {
  if (agent.state.status === 'idle') {
    cleanup();
    resolve({ speech, toolCallsCount });
  }
  break;
}
```

With `reset()` happening *after* the trace, `agent.state.status` was still `'streaming'` at emit time. The check failed, cleanup never ran, the agent stayed in `this.agents` indefinitely.

PR #32 restores correctness, but it cements a contract that was never advertised: the next "harmless" trace refactor (e.g. emitting earlier so the UI updates before a slow `dispatchSpeech`) will silently break ephemeral cleanup again. We just pinned the emit order without saying so.

## Proposed fix

Promote ephemeral-stream completion **off** the trace bus.

**Preferred — promise returned by the stream driver.** `driveStream` (or a wrapper) returns `Promise<{stopReason, speech, toolCallsCount, ...}>` that resolves when the stream truly settles, sourced from the state machine. `runEphemeralToCompletion` awaits that promise; no trace listener, no agent-name filter, no state-at-emit check. Test surface collapses to "mock the promise"; current tests must choreograph trace events in the right order with the right state.

**Acceptable — structured process event.** Push `process:agent-settled` onto the process queue alongside other `ProcessEvent`s, with `{agentName, stopReason, ...}`. Less elegant (broadcast-filter-by-name re-emerges) but uses the explicit business-logic transport.

**Avoid:** emitting a *second* trace dedicated to ephemeral cleanup. That just relocates the violation.

## Related sites to audit

Grep `this.onTrace(` outside `framework.ts`'s own logging/UI/CLI consumers. Anything in a control-flow path is a candidate:

- `errorPolicy.onInferenceError` retry bookkeeping in `driveStream` reads `attempt`/`requestId` off the same flow.
- Any module that reacts to `inference:completed` to do persistent bookkeeping (lessons-on-completion writes, eventgate state, usage tracking side effects).

Not all of these are necessarily wrong — observability *can* legitimately produce side effects (logging, metrics). The question is whether the framework's own correctness depends on the trace firing. Where it does, the trace listener should be replaced with a state-machine signal.

## Acceptance criteria

- [ ] `runEphemeralToCompletion` no longer subscribes to trace events
- [ ] Subagent tests pass without choreographing trace emit order or state-at-emit
- [ ] Document in framework README / CONTRIBUTING: traces are observability only; framework-internal control flow must not consume them
- [ ] Optional: lint or guard test that fails if framework-internal files (outside `trace.ts` / `framework.ts` emit sites) call `this.onTrace(` for control-flow purposes

## Context

Discovered while postmortem-ing a duplicate-Zulip-reply incident in a triumvirate fleet deployment. The zombie subagent burned ~half the run's input tokens (2.28M input / 21k output across 18 inferences) after its parent-side promise had already settled.

mvanhorn:
## Summary

`runEphemeralToCompletion` now settles from a framework-owned signal resolved by the stream driver, instead of subscribing to `inference:completed` / `inference:turn_ended` / `inference:exhausted` on the trace bus. Trace emit order is no longer load-bearing for ephemeral control flow.

## Why this matters

Traces are documented as observability-only (`src/types/trace.ts`), but `runEphemeralToCompletion` consumed them to resolve its caller's promise and deregister the agent (`this.onTrace(...)`). That made trace emit order and state-at-emit an undocumented API surface: a past emit-order refactor produced a zombie subagent that kept running, which PR #32 mitigated by pinning the emit order in place rather than removing the dependency. Issue #33 asks for the settle to come from the state machine instead.

## Changes

- Added `agentSettles: Map<string, Deferred<AgentSettleResult>>` on `Framework` and an `AgentSettleResult` type (`src/types/framework.ts`).
- The stream driver resolves the deferred at each terminal transition — `complete`, `turn_ended` (via an `endTurn` tool result), and every `exhausted` path — and never on retry branches, so the framework retry loop keeps the deferred alive.
- `runEphemeralToCompletion` awaits that promise with the 30s startup and 15-min idle watchdogs racing against it, then cleans up `agents` and the deferred in a `finally`. It registers no trace listener.
- README now states that traces are observability-only and framework-internal control flow must not consume them.

## Testing

`test/ephemeral-settle.test.ts` covers plain-text completion, tool rounds with final speech + call count, the `endTurn` tool-result path, retry exhaustion (rejects; intermediate retries do not settle), emit-order independence (asserts no trace listener is registered), the startup watchdog, and a source guard that fails if `this.onTrace(` reappears in framework-internal control flow.

- `npm run typecheck`: passes
- `npm run build`: passes
- New suite: 7/7 pass
- Full suite: 227/230 (the 3 failures are pre-existing MCPL stdio-transport tests on `main`, unrelated to this change)

Closes #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,11 @@ framework.onTrace((event) => {
 });
 ```
 
+Trace events are observability-only. They are intended for logs, metrics, UI
+updates, and debugging, not framework-internal control flow. Internal lifecycle
+code should use state-machine signals or process events instead of consuming
+traces, so trace emit order does not become an API contract.
+
 Query raw inference logs:
 
 ```typescript
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -31,6 +31,7 @@ import type {
   ProcessEvent,
   EventResponse,
   ModuleProcessResponse,
+  AgentSettleResult,
   ToolCall,
   ToolCallEvent,
   ToolResult,
@@ -219,6 +220,18 @@ interface RewindRecord {
   discordRef?: { channelId: string; messageId: string };
 }
 
+interface Deferred<T> {
+  promise: Promise<T>;
+  resolve: (value: T) => void;
+  reject: (reason?: unknown) => void;
+}
+
+interface AgentSettleProgress {
+  inferenceStarted: boolean;
+  lastActivity: number;
+  toolCallsCount: number;
+}
+
 export class AgentFramework {
   private store: JsStore;
   private ownsStore: boolean;
@@ -248,6 +261,9 @@ export class AgentFramework {
   private processLoggingBroadcast: boolean;
   private activeStreams: Map<string, Promise<void>> = new Map();
   private pendingAssistantBlocks: Map<string, ContentBlock[]> = new Map();
+  private turnEndedStreams: Set<string> = new Set();
+  private agentSettles: Map<string, Deferred<AgentSettleResult>> = new Map();
+  private agentSettleProgress: Map<string, AgentSettleProgress> = new Map();
   /** Per-agent count of consecutive exhausted inferences (reset on any success).
    *  Drives hard-down escalation — see noteInferenceExhausted. */
   private consecutiveInferenceFailures: Map<string, number> = new Map();
@@ -956,6 +972,47 @@ export class AgentFramework {
     return { agent, contextManager, cleanup };
   }
 
+  private createDeferred<T>(): Deferred<T> {
+    let resolve!: (value: T) => void;
+    let reject!: (reason?: unknown) => void;
+    const promise = new Promise<T>((res, rej) => {
+      resolve = res;
+      reject = rej;
+    });
+    return { promise, resolve, reject };
+  }
+
+  private touchAgentSettle(agentName: string, inferenceStarted = false): void {
+    const progress = this.agentSettleProgress.get(agentName);
+    if (!progress) return;
+    if (inferenceStarted) {
+      progress.inferenceStarted = true;
+    }
+    progress.lastActivity = Date.now();
+  }
+
+  private recordAgentSettleToolCalls(agentName: string, count: number): void {
+    const progress = this.agentSettleProgress.get(agentName);
+    if (!progress) return;
+    progress.toolCallsCount += count;
+    progress.lastActivity = Date.now();
+  }
+
+  private getAgentSettleToolCalls(agentName: string): number {
+    return this.agentSettleProgress.get(agentName)?.toolCallsCount ?? 0;
+  }
+
+  private settleAgent(agentName: string, result: AgentSettleResult): void {
+    const settle = this.agentSettles.get(agentName);
+    if (!settle) return;
+    this.touchAgentSettle(agentName);
+    if (result.stopReason === 'exhausted') {
+      settle.reject(new Error(result.error ?? 'Unknown error'));
+      return;
+    }
+    settle.resolve(result);
+  }
+
   /**
    * Run an ephemeral agent to completion through the framework's event loop.
    *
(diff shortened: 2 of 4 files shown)


Anarchid (changes requested):
Thanks — this is the right fix for #33, and well-shaped: the settle signal comes from the state machine instead of the trace bus, the watchdogs survive the refactor, and the test suite pins the things that actually broke in production (retry keep-alive, emit-order independence, the `this.onTrace(` source guard). We want to merge this. Requesting changes on the following first.

### 1. Blocker — the `aborted` early-return drops the EventGate clear for endTurn (wake-wedge on this branch as submitted)

```ts
case 'aborted': {
  if (this.turnEndedStreams.delete(`${agent.name}:${myStreamId}`)) {
    return;
  }
```

On your base (590b739), the endTurn flow was: `cancelStream()` → membrane delivers an `aborted` event → the handler ran `agent.reset()`, emitted `inference:exhausted`, and — critically — called `this.eventGate?.onInferenceEnded(agent.name)`. Your early `return` skips that call, and neither the endTurn branch in `handleProcessEvent` nor (on this branch) `driveStream`'s `finally` clears the gate. With an EventGate configured, one endTurn tool call leaves the agent permanently in the gate's `inferring` set, which buffers all subsequent events — the framework silently stops waking. That's the exact wedge fixed on main in 9948607 (Jul 8) by adding `onInferenceEnded` to `driveStream`'s `finally`. Your tests don't configure a gate, so they can't observe it.

After a rebase the `finally` masks this, but please still add an explicit `this.eventGate?.onInferenceEnded(agent.name)` in the endTurn branch next to `agent.reset()` — every other reset site pairs them, and the abort path's correctness shouldn't silently depend on a backstop in a commit this branch wasn't written against.

### 2. Blocker — rebase onto current main (≥ d34e2c7)

main has moved by four commits that interact with this diff: 9948607 (the gate `finally`, load-bearing per item 1), d453165 + 6e6380b (failure observability and its test fix), d34e2c7 (`pretest` fixture copy). The merge is textually clean — we ran it — but we want the suite green on the actual merge target. Side effect: the "3 pre-existing MCPL stdio failures" in your description were a fixture-packaging gap, since fixed by the `pretest` script; after rebasing you should see a fully green run, and the description should be updated to match.

### 3. Blocker — the suppressed spurious `inference:exhausted` is an observable trace change; document it

The `turnEndedStreams` guard means endTurn is no longer followed by the bogus `inference:exhausted` that main emits today. We agree it's spurious and suppressing it is correct — but consumers exist that learned to treat that event as the terminal signal for an endTurn'd agent (we have one in-house with a pinned test; we'll fix our side). Given this PR's whole thesis is that trace behavior becomes an accidental API surface, the trace behavior it changes needs an explicit callout in the PR description rather than being discovered downstream.

### 4. Extend the cancel-guard to the budget-restart path

```ts
} else if (overBudget) {
  agent.cancelStream();
```

Identical race shape to the one you just fixed: `cancelStream()` delivers `aborted` before the restart bumps `streamId`, so `agent.streamId === myStreamId` still holds and `settleAgent(..., 'exhausted')` fires — an ephemeral that merely crosses its context budget mid-run gets its promise rejected while the framework is restarting the stream. This hole predates the PR (the old trace listener rejected on the same event), but the PR builds the mechanism that fixes it for endTurn and leaves the sibling three lines down. If the guard keys on "framework-initiated, non-terminal cancel" rather than "endTurn," both paths are covered.

### 5. Consolidate the per-run state

`agentSettles` + `agentSettleProgress` + `turnEndedStreams` plus four accessor methods are three parallel per-agent structures that must be created and torn down in lockstep. A single `Map<string, EphemeralRun>` (deferred + progress + cancel flag) removes the desync risk. Related cleanup:

```ts
toolCallsCount: this.agentSettleProgress.has(agent.name)
  ? this.getAgentSettleToolCalls(agent.name)
  : toolCallsCount,
```

The `: toolCallsCount` branch is unreachable — `settleAgent` no-ops unless the settle entry exists, and settle/progress entries are created and deleted together — and the local `toolCallsCount` counter exists only to feed it (and would disagree with the progress counter across stream restarts if it were ever live). The endTurn settle site already skips the ternary. One counter, please.

### 6. Revert the whitespace-only reindent

`this.emitTrace({ type: 'process:completed', ... })` picked up two extra spaces of indentation with no functional change.

### 7. Tests (fine as a fast-follow, but these will bite)

- The startup-watchdog test replaces the global `setTimeout`/`setInterval`/`clear*` and identifies the watchdog by `delay === 30_000`. If `STARTUP_TIMEOUT_MS` ever changes, the test hangs rather than fails, with every other timer broken during the patched window. Injecting the timeouts (an options parameter) or using `node:test` mock timers would make it robust.
- The guard test's allowlist is an exact source-line string; any reformat of that line fires a false positive, and a guard that cries wolf gets deleted. A structural match (file + `onTrace: (listener) =>` pattern) or a comment-marker annotation would hold up better.

Happy to merge once 1–4 are addressed (5–7 negotiable). If you don't plan to iterate on this, let us know and we'll adopt the branch and take it the rest of the way ourselves.

mvanhorn:
Addressed in 25e16b2: the aborted turn-ended early-return path now also releases the event gate (onInferenceEnded), so an abort during a tool-only turn can't leave the gate held. Verified locally: the ephemeral-settle suite passes (7 tests); the mcpl-reconnect tests fail identically with and without this change on my machine (child process handshake timing), so those look environmental rather than related.

Anarchid:
Adopted and merged via #63 — thank you for this work, and for the quick 25e16b2 follow-up on the gate release; both of your commits landed as-authored, so the refactor is yours in the history.

We took it the rest of the way per the review (merge with current main, budget-restart cancel-guard, per-run state consolidation, test hardening) after four days of quiet and because our downstream schedule needed it — the maintainer-edit push to your branch was refused by GitHub (403 despite maintainer-edits being enabled), hence the superseding PR rather than commits here. Details of what changed on top are in #63's description.

Closing in favor of #63. Closes #33 either way — thanks again!

PR #57, opened 2026-07-09
fix: eliminate set-rewrite quadratic states (inference log, process log, turn checkpoints) + whole-store scaling gate

Anarchid:
## What

Three framework states were persisted by rewriting their full accumulated value on every update, making each record O(history) and aggregate disk O(n²) in update count. All were found by retro-benchmarking evacuated production stores (log-log slope of bytes-per-record vs update index; flat ≈ healthy, ≈1 ≈ quadratic aggregate):

| state | prod evidence | fix |
|---|---|---|
| `framework/inference-log` | slope **0.98** in three prod stores — registered `append_log`, but the writer did `getStateJson → push → setStateJson`, defeating the strategy with a whole-array `Set` per inference | one `appendToStateJson` per entry |
| `framework/process-log` | identical bug, same writer pattern | identical fix |
| `framework/turn-checkpoints` | slope **0.86–0.94** — one `Record<agentName, list>` map rewritten on every turn of every agent; keys are per-spawn-unique subagent names and were never evicted, so the map grew without bound in agents-ever-spawned | one Tree state (`framework/turn-checkpoints/tree`) keyed by agent name — per-key O(entry) writes, keys **removed** on disposal, exactly one state registration for the life of the store (a slot-per-agent design would leak a permanent registration per spawn into the state index, which is rewritten + fsynced every sync tick); legacy map kept as read-only fallback, migrated per agent on first save, with eviction tombstones so a reused name can't inherit a dead agent's checkpoints |

Plus the enabler: `ModuleContext` only offered full-object `setState`, which funnels modules into the same pattern (a downstream module's corpus measured **93.4 MB — a third of a 278 MB prod store**). New granular log-state API: `registerLogState` / `appendToLog` / `editLogItem` / `getLog` / `getLogLength`, backed by chronicle's `append_log` primitives, with a registered-name guard — chronicle accepts appends to unregistered states but never snapshots them, so a forgotten `registerLogState` or typo'd name is an immediate error instead of a silent full-chain-replay-on-restore trap. `registerState` catches rethrow anything that isn't the typed "State already exists". Agent disposal also clears the per-agent `lastInferenceAt` / `staleWarnAt` diagnostics maps, which had the same unbounded-in-agent-names shape (in-memory only).

## Compatibility

- Appending after a legacy `Set`-written array is defined chronicle behavior — pinned by a fixture-style test (open old-layout store, append new way, read back).
- Old checkpoint maps stay readable through the fallback path; each agent's first new checkpoint migrates its list into its own slot.
- `runUntilIdle()` now also drains `pendingRequests` — direct inference requests (`runEphemeralToCompletion`) bypass the event queue, and the helper could previously return while one was still waiting.

## Tests

- `test/state-scaling.test.ts` — unit gates: Nth record size independent of N for each fixed writer, plus both legacy-store compat paths.
- `test/state-scaling-e2e.test.ts` — regime gate: drives a synthetic fleet workload through the public surface (40 external-message turns with tool rounds, 12 ephemeral spawn-and-dispose cycles, process logging on), then sweeps **every** state family in the store and asserts none grows superlinearly (slope < 0.3; capped families get a hard max-record-size bound). A quadratic writer added anywhere fails CI the day it lands, including in states this PR has never heard of.

Suite: 254/254 green on top of current main (post-#54). The e2e sweep also asserts exactly one checkpoint state is ever registered, and lists (rather than silently skips) families below its slope-fit threshold.

## Before / after (identical synthetic workload, pre-fix build vs this branch)

Totals are the whole store directory on disk — records, content-addressed blobs, and the state index all count, so the tree-backed checkpoints can't hide bytes in blob storage. Both runs finish with the same 31 registered states (a slot-per-agent design would have left 44).

| family (state-update records) | before | after | |
|---|---|---|---|
| `framework/inference-log` | 17.36 MB | 0.51 MB | **34×** — last record alone: 499 KB → 2.2 KB |
| `framework/process-log` | 7.31 MB | 0.24 MB | **31×** |
| `framework/turn-checkpoints` | 368 KB | 18 KB | tree ops; checkpoint payloads in blobs, counted in the total below |
| `messages` (control, untouched) | 192 KB | 192 KB | within 0.1% |
| **whole store on disk** | **25.5 MB** | **1.28 MB** | **19.8×** |

The gap widens with session length — "before" grows per-record, "after" doesn't.

📉 **Interactive charts** (bytes-per-record trajectories per family, incl. the control): **[rendered page](https://gistcdn.githack.com/Anarchid/5f80439539f96db9b0e036db6912972e/raw/quadratic-states-before-after-standalone.html)** · [source gist](https://gist.github.com/Anarchid/5f80439539f96db9b0e036db6912972e)

## Notes for review

- First commit (`1d89a96`, abort/catch-path inference-log persistence) is an older standalone fix riding along on this branch — it adds the write sites that make the append fix more valuable, but reverts independently if you want it split out.
- `logs/` added to `.gitignore`: the new `failures.log` append (from the diagnostics work on main) is cwd-relative, so `npm test` drops `logs/failures.log` in the repo root. Might be worth rooting it under a configured path in a follow-up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ data/
 .env
 package-lock.json
 bun.lock
+logs/
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   "author": "Anima Research",
   "license": "MIT",
   "dependencies": {
-    "@animalabs/chronicle": "^0.2.0",
+    "@animalabs/chronicle": "^0.2.2",
     "@animalabs/context-manager": "^0.5.5",
     "@animalabs/membrane": "^0.5.64",
     "chokidar": "^4.0.3",
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -3708,11 +3708,27 @@ export class AgentFramework {
             // Only reset if this is still the active stream (a budget restart
             // may have already started a new stream, bumping streamId)
             if (agent.streamId === myStreamId) {
+              const durationMs = Date.now() - startTime;
               agent.reset();
               this.emitTrace({
                 type: 'inference:exhausted',
                 agentName: agent.name,
                 error: `Stream aborted: ${reason}`,
+              });
+              // Postmortem 2026-05-28 P2 #7: persist the abort to the
+              // inference log so future investigations can attribute the
+              // terminal cause without relying on live in-memory reducer
+              // state. Without this, abort-terminated inferences are
+              // invisible to forensic queries (only request-side telemetry
+              // via llm-calls.jsonl shows them, and only by absence).
+              this.logInference({
+                timestamp: startTime,
+                agentName: agent.name,
+                requestId,
+                success: false,
+                error: `Stream aborted: ${reason}`,
+                request: compiledRequest ?? { note: 'streaming request aborted' },
+                durationMs,
               });
               this.eventGate?.onInferenceEnded(agent.name);
             }
@@ -3738,6 +3754,7 @@ export class AgentFramework {
       // Stream itself threw (unexpected) — no retry path here, so also emit
       // inference:exhausted so ephemeral agent promises can settle.
       const err = error instanceof Error ? error : new Error(String(error));
+      const durationMs = Date.now() - startTime;
       this.emitTrace({
         type: 'inference:failed',
         agentName: agent.name,
@@ -3748,6 +3765,18 @@ export class AgentFramework {
         type: 'inference:exhausted',
         agentName: agent.name,
         error: err.message,
+      });
+      // Postmortem 2026-05-28 P2 #7: catch-path failures (stream itself
+      // threw) were previously only visible via in-memory trace listeners.
+      // Persist to inference log for forensic attribution.
+      this.logInference({
+        timestamp: startTime,
+        agentName: agent.name,
+        requestId,
+        success: false,
+        error: `Stream threw: ${err.message}`,
+        request: compiledRequest ?? { note: 'streaming request threw' },
+        durationMs,
       });
       agent.reset();
       this.eventGate?.onInferenceEnded(agent.name);
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -43,7 +43,7 @@ import type {
 } from './types/index.js';
 import { ProcessQueueImpl } from './queue.js';
 import { Agent } from './agent.js';
-import { ModuleRegistry } from './module-registry.js';
+import { ModuleRegistry, isStateExistsError } from './module-registry.js';
 import { McplServerRegistry } from './mcpl/server-registry.js';
 import { FeatureSetManager } from './mcpl/feature-set-manager.js';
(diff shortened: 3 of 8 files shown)


Anarchid:
All seven findings addressed, in `00485b0` + `fee96a7` (branch re-benchmarked, PR body updated to match). Point by point:

**1 — chronicle floor**: bumped to `^0.2.2` (`00485b0`). Went past 0.2.1 for the reason you gave: `cde8ecb`'s per-item cache is what makes `editLogItem`'s advertised point-read behavior real, and the tree NAPI (used below) is 0.2.2 as well.

**2 — index residue**: you're right, and "removed" was wrong — it was relocated to a file the sweep doesn't read. Took the ideal outcome: checkpoints now live in a single **Tree state** (`framework/turn-checkpoints/tree`), one key per agent pointing at a JSON blob of its ≤20-checkpoint list. Per-key O(entry) writes, `treeRemove` on disposal, exactly one registration for the life of the store. The e2e sweep now asserts `listStates()` contains exactly one checkpoint state, and the re-run benchmark confirms both runs end at the same 31 registered states (slots would have left 44). The benchmark also switched to honest accounting — whole store directory on disk, blobs and index included — so the tree can't hide bytes where the old metric didn't look: 25.5 MB → 1.28 MB (19.8×).

**3 — eviction registers state for the never-checkpointed**: gone structurally — nothing registers per-agent state anymore; the one tree registration happens in `create()`.

**4 — unregistered appendToLog**: `ModuleContextImpl` now tracks registered names and throws from all four log methods on an unknown name, with the replay-trap rationale in the error message. Pinned by test (`appendToLog('lgo', …)` → `/not registered/`).

**5 — catch-all labeled "Already registered"**: both sites now rethrow anything that isn't the typed `State already exists` message. Agreed on the wish diagnosis.

**6 — legacy-map ghosts**: eviction now tombstones (empty list shadowing the fallback) when the agent's checkpoints live only in the legacy map, and only `treeRemove`s otherwise. Unit test covers the ghost case and evict idempotence (second evict writes no new record).

**7 — silent sub-threshold skip**: the sweep prints the families it can't fit, count and names. Fair hit — it was exactly the cap-without-logging my own harness rules prohibit.

Benchmark page updated in place (same links): [rendered](https://gistcdn.githack.com/Anarchid/5f80439539f96db9b0e036db6912972e/raw/quadratic-states-before-after-standalone.html) · [gist](https://gist.github.com/Anarchid/5f80439539f96db9b0e036db6912972e). The turn-checkpoints chart is the one worth re-looking at: the after-series is now flat ~200 B tree ops instead of the capped ramp.

One honest caveat on #2's residual: ephemeral agents still leave `subagent/<name>/messages` context-manager states behind (pre-existing, outside this PR's writers), so the state index still grows with spawns-ever via that path. Filed mentally for the campaign's CM phase — happy to open an issue if you want it tracked upstream.

PR #58, opened 2026-07-10
Compile robustness: empty-text 400 sanitize + OverBudget deadlock breaker

LariTesserae:
Two independent fixes for ways an agent goes silently mute, both field-observed on a resident agent on 2026-07-10 (same incident day, full timeline available on request).

## 1. Strip empty text blocks from live activation requests (`agent.ts`)

The Anthropic API rejects empty text blocks with 400 `messages: text content blocks must be non-empty`. Empty blocks occur naturally in stored history (tool-only turns, delivery-failure placeholders, silent turns); one such block in the compiled window mutes the agent entirely — `[inference-failed]` on every attempt until the message ages out. Twin of context-manager's `stripEmptyTextBlocks` on the compression path. Sanitize runs before the trailing-assistant check so dropping a trailing message can't leave an assistant message last.

## 2. OverBudget deadlock breaker (`framework.ts`)

When compile throws `OverBudgetError`, the compression drain never runs: it is driven by successful activity, which the over-budget state prevents — a closed loop with no internal exit. Field data: 36 minutes hard-down, zero self-rescue, external operator intervention required. Fix: on an OverBudget-class inference failure, kick the strategy drain directly (bounded 8 ticks, best-effort, never throws) so folding/merging frees space for the next compile.

## Verification
- Both patches ran in production on a live resident agent for several hours (they are ports of the battle-tested local patches from the incident).
- `bun test`: failure count identical to main baseline (9), no new failures.

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -336,6 +336,26 @@ export class Agent {
 
     let { messages, systemInjections } = await this.compileWithInjections(budget, injections);
 
+    // Sanitize: strip empty/whitespace text blocks and drop messages left with
+    // no content. The Anthropic API rejects empty text blocks with 400
+    // "messages: text content blocks must be non-empty"; when such a block
+    // reaches a live activation request the agent cannot complete ANY turn
+    // ([inference-failed] on every attempt) until the offending message ages
+    // out of the window. Empty text blocks occur naturally: tool-only turns,
+    // delivery-failure placeholders, silent/skip turns. Non-text blocks pass
+    // through unchanged, so tool pairing is unaffected. (Field-observed
+    // 2026-07-10 on a resident agent: one empty block muted the agent's live
+    // path entirely; twin of context-manager's stripEmptyTextBlocks on the
+    // compression path.)
+    messages = messages
+      .map((m) => ({
+        ...m,
+        content: m.content.filter(
+          (b: any) => !(b.type === "text" && (typeof b.text !== "string" || b.text.trim() === "")),
+        ),
+      }))
+      .filter((m) => m.content.length > 0);
+
     // Safety: ensure messages don't end with an assistant message.
     // Some models reject trailing assistant messages ("prefill not supported"),
     // and after context compression a stale assistant turn can end up last.
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -355,7 +355,9 @@ export class Agent {
       .map((m) => ({
         ...m,
         content: m.content.filter(
-          (b: any) => !(b.type === "text" && (typeof b.text !== "string" || b.text.trim() === "")),
+          // typeof guard is deliberate runtime defense: history loaded from
+          // disk can carry a non-string `text` despite what the types claim.
+          (b: ContentBlock) => !(b.type === "text" && (typeof b.text !== "string" || b.text.trim() === "")),
         ),
       }))
       .filter((m) => m.content.length > 0);
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -4304,6 +4304,29 @@ export class AgentFramework {
       }
     }
 
+    // (2b) OverBudget deadlock breaker. When compile fails with
+    // OverBudgetError ("...exceed hard budget..."), the normal compression
+    // drain never runs: it is driven by successful activity, which the
+    // over-budget state prevents - a closed loop with no internal exit.
+    // Field data (2026-07-10, resident agent): 36 minutes hard-down, zero
+    // self-rescue; only an operator raising the budget externally could
+    // break the loop. Break it here instead: kick the strategy drain
+    // directly so folding/merging frees space for the next compile.
+    // Bounded ticks, best-effort, never throws.
+    if (agent && /exceed hard budget|OverBudget/i.test(reason)) {
+      (async () => {
+        try {
+          const cm = agent.getContextManager();
+          for (let i = 0; i < 8; i++) {
+            await cm.tick();
+          }
+          console.error(`[inference-failed] drain kicked for ${agentName} (OverBudget breaker, 8 ticks)`);
+        } catch (err) {
+          console.error(`[inference-failed] drain kick failed for ${agentName}:`, err);
+        }
+      })();
+    }
+
     // (3) Hard-down escalation on repeated identical failure.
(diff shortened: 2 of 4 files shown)


Anarchid (commented):
Reviewed both patches with a local checkout: `tsc --noEmit` clean; full test suite shows **zero new failures** vs the base commit (the state-scaling gate failures are identical on `ceda16a` and PR head, matching your bun baseline observation). Both fixes address real failure modes and the placement choices are mostly sound — the sanitize running *before* the trailing-assistant check is the right ordering, and dropping whole messages is safe because membrane's wire-boundary net (`normalizeToolPairs` + `mergeConsecutiveRoles`) collapses any same-role neighbours it creates downstream.

That said, two things I'd like addressed before merge, and a few follow-ups.

## Before merge

### 1. The OverBudget trigger is a regex over error-message prose (`framework.ts:4316`)

```ts
if (agent && /exceed hard budget|OverBudget/i.test(reason)) {
```

`reason` is `err.message`, so the match works only because `OverBudgetError`'s constructor in context-manager (`adaptive/picker.ts:58`) happens to interpolate "exceed hard budget" into its message. The `|OverBudget` alternative is dead weight — the class name appears in `err.toString()`, not `err.message`. If CM ever rewords that message (it has no indication it's load-bearing for a breaker in a different package), the breaker dies **silently** — which is worse than not having it, because everyone will believe the net exists.

`OverBudgetError` isn't exported from CM's package root, so a clean `instanceof` wasn't available to you — but `err.name` is set to `'OverBudgetError'` (picker.ts:65) and survives the package boundary without imports. Suggest threading it through the existing classification at the two `inference:exhausted` emit sites:

```ts
errorType: err instanceof MembraneError ? err.type
  : err.name === 'OverBudgetError' ? 'over_budget'
  : undefined,
```

and gating (2b) on `errorType === 'over_budget'` (keep the message regex as a fallback if you like, but the name check should be primary).

### 2. No regression tests — for a PR whose theme is "this keeps happening in the field"

Both fixes are cheaply testable:

- the sanitize via `buildActivationRequest` with a compile stub returning empty text blocks — the method is documented as pure/non-mutating and safe to call directly;
- the breaker via `noteInferenceExhausted` with a mock agent counting `tick()` calls.

"Ran in production for several hours" is good operational evidence, but the prose-regex above can only rot *silently* because no test pins it. Two small tests close that hole.

## Smaller points (fix now or accept as noted)

- **`(b: any)` at `agent.ts:354`** — `m.content` is `ContentBlock[]`, a discriminated union; `b.type === "text"` narrows fine and the predicate typechecks without the `any`. The `typeof b.text !== "string"` runtime check is reasonable defense against untyped history from disk, but with `any` you can't tell deliberate defense from silencing — type it as `ContentBlock` and let the `typeof` stand as visibly intentional.
- **Overlapping drain kicks** — the IIFE fires on *every* matching failure, and the scenario it exists for is "every activation fails, repeatedly." Concurrent kicks are safe (strategy `tick()` self-guards via `pendingCompression`) but ticks landing while another kick's compression is in flight return immediately having done nothing — so "8 ticks" isn't 8 units of drain work, and the success log prints `(OverBudget breaker, 8 ticks)` even when every tick no-oped. A per-agent in-flight flag would make behaviour deterministic and the log honest.

## Follow-ups (not blocking, will file separately unless you want them)

- **The sanitize protects one of two request-assembly paths.** `buildActivationRequest`'s docstring calls it "the single source of truth for request assembly," but `runInferenceWithInjections` (agent.ts:202–221) assembles its own request and doesn't get the sanitize. No live callers today, but it's public API — hoisting the sanitize into a shared helper (or routing `runInference*` through `buildActivationRequest`) would make the docstring true.
- **This is the third copy of this logic.** The PR calls itself a "twin of context-manager's `stripEmptyTextBlocks`" — twins are the symptom. Membrane already owns a wire-boundary repair pass for exactly this bug class (orphan tool_use, same-role envelopes, the adapter-level block sanitize in `buildRequest`); empty-text stripping belongs there, covering every consumer and every path at once, instead of one path per incident per repo.
- **CM should export `OverBudgetError` from its root** — it's part of the cross-package behavioral surface now, whether CM knows it or not.

Good field data, correct diagnoses, honest scoping — the 36-minute-outage class is worth killing. Address #1 and #2 and this merges.

🤖 Review by [Claude Code](https://claude.com/claude-code)

Anarchid:
Both blocking items addressed in `4339d9c` (after a clean merge of main/0.6.4 in `8c026d8`); full suite green at 284/284.

**1 — prose-regex trigger:** Took your suggestion, with one refinement: instead of inlining the ternary at each emit site, classification now lives in a single `classifyInferenceError(err)` helper (MembraneError → its typed classification; `err.name === 'OverBudgetError'` → `'over_budget'`), used at **all three** `inference:exhausted` emit sites — including the stream-threw catch path, which previously emitted unclassified and would have bypassed both breakers. The drain kick gates on `errorType === 'over_budget'`; `/exceed hard budget/i` stays as a fallback for reason strings that lost their Error object, with a comment stating explicitly that it may rot and the errorType gate is the load-bearing one. The dead `|OverBudget` alternative is gone.

**2 — tests:** `overbudget-drain-breaker.test.ts` pins the classification with a deliberately reworded message (so a CM wording change fails the suite rather than silently killing the breaker), plus: kick on `over_budget`, prose fallback, non-triggering error types (`server`/`invalid_request`/unclassified), overlap suppression, and flag-clearing on a throwing tick. `activation-request-sanitize.test.ts` covers empty/whitespace/non-string text, emptied-message dropping, tool-block passthrough, and the sanitize→trailing-assistant ordering you called out as correct.

**Smaller points:** both taken. `(b: ContentBlock)` with a comment marking the `typeof` guard as deliberate disk-history defense; overlapping kicks suppressed via a per-agent in-flight set (cleared in `finally`), and the success/failure logs now print the *actual* completed tick count.

**Follow-ups:** agreed on all three — `runInferenceWithInjections` not getting the sanitize, membrane as the right home for wire-boundary empty-text stripping, and CM exporting `OverBudgetError` (which would upgrade the name check to a real `instanceof`). Filing them as issues rather than growing this branch.

PR #66, opened 2026-07-16
fix: prevent deferred inference requests from starving the event loop

antra-tess:
## Summary

- always yield a real event-loop turn when inference requests are deferred behind a busy agent
- preserve low-latency processing when queue traffic and background work coexist
- throttle stale-request traces together with their once-per-minute warning
- add regression coverage for the exact scheduler states observed during the Sol outage

## Root cause

When a new wake arrived while an agent was `streaming` or `waiting_for_tools`, `processInferenceRequests()` requeued it. `processNextEvent()` treated the non-empty request queue as progress and skipped its normal backoff. If `activeStreams` was empty or stale at the same time, the loop had no macrotask yield at all.

That microtask-only polling loop pinned Bun at 100% CPU and starved the tool-result and HTTP I/O needed to make progress. Sol remained `waiting_for_tools`, while Fleet health probes timed out and reported the agent down.

## Impact

Deferred wakes can no longer turn an otherwise recoverable tool wait into a self-sustaining event-loop wedge. Tool results, Web UI health checks, Discord traffic, and timers continue receiving event-loop time while the wake remains queued.

## Validation

- `npm run build`
- focused scheduler, reconnect, and channel-routing tests
- full agent-framework suite: 326 passed, 0 failed
- deployed to Sol as a canary: Fleet returned `UP`, Web UI HTTP 200, zero pending requests, and normal CPU

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -322,7 +322,7 @@ export class AgentFramework {
   private maintenanceRunId = 0;
   private currentMaintenanceRun: ContextMaintenanceRun | null = null;
   private maintenanceHistory: ContextMaintenanceRun[] = [];
-  /** Last time we console-warned about stale (busy-requeued) inference requests, per agent. */
+  /** Last time we reported stale (busy-requeued) inference requests, per agent. */
   private staleWarnAt = new Map<string, number>();
   /** Per-agent last inference activity (epoch ms), for /healthz + doctor tooling. */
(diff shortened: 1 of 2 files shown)


antra-tess:
Closing as superseded: this landed on main verbatim as dff0bef6 (2026-07-17, "fix: yield while inference requests are deferred") the day after this draft opened — identical footprint including the busy-agent-scheduler test. Main has since evolved the landed code past this branch (turn-alive guard e086cc18, budget-restart semantics #117), so merging the branch now would regress those. Nothing left to land here.

PR #74, opened 2026-07-31
fix: Attribute tool-call state across multiple feature sets

### LINKED ISSUE #48: MCPL host state misattributed when a server declares >1 stateful feature set
## Summary

MCPL host-managed state is misattributed when a server declares **more than one stateful feature set**. The host injects and records state for *the first* stateful feature set only, regardless of which feature set the called tool actually belongs to.

## Where

`CheckpointManager.getStatefulFeatureSet(serverId)` returns the **first** stateful feature set (Map iteration order):

```ts
// src/mcpl/checkpoint-manager.ts
getStatefulFeatureSet(serverId: string): string | null {
  const prefix = `${serverId}:`;
  for (const [key] of this.trees) {
    if (key.startsWith(prefix)) return key.slice(prefix.length);  // first only
  }
  return null;
}
```

It's then used for **both** sides of the tool-call state exchange in `src/framework.ts`:

```ts
// inject (≈3524)
const fs = this.checkpointManager.getStatefulFeatureSet(serverId);
if (fs) { stateParams = isHostManaged ? { state: getCurrentState(serverId, fs) }
                                      : { checkpoint: getCurrentCheckpoint(serverId, fs) }; }

// record (≈3542)
if (result.state) {
  const fs = this.checkpointManager.getStatefulFeatureSet(serverId);
  if (fs) this.checkpointManager.recordCheckpoint(serverId, fs, result.state);
}
```

## Consequence

For a server with multiple stateful feature sets (e.g. xgate/x-mcpl declares `x.post`, `x.notifications`, `x.feed`, `x.dm`):

- A checkpoint returned by a tool belonging to feature set **B** is recorded onto feature set **A** (the first one) — corrupting A's tree and silently losing B's state.
- Every tool call injects **A**'s state/checkpoint, so tools in B never receive their own feature set's state.

Only the first stateful set works; the rest are broken on the tool-call path.

## Root cause

The wire gives the host no way to attribute a tool call (or a `result.state`) to a feature set:
- `tools/list` tool definitions carry no `featureSet`.
- The `result.state` checkpoint carries no `featureSet`.

So `getStatefulFeatureSet` "guesses" the first set. That's only correct when a server has exactly one stateful feature set.

Note: the **server-driven** path (`state/update`, which carries an explicit `featureSet`) is unambiguous and not affected — only the synchronous tool-result piggyback + pre-call injection path is.

## Suggested fix

Attribute state by an **explicit signal only**, and never guess among several:

1. If `result.state` carries an explicit `featureSet`, use it.
2. Else if tools are tagged (`featureSet` / `_meta.featureSet`), map tool → set.
3. Else if exactly one stateful set exists (optionally narrowed by host- vs server-managed mode), use it.
4. Otherwise **log a warning and drop** the update — do not record/inject against the wrong set.

This is backward-compatible (single-stateful-set servers keep working) and "lights up" per-set correctness once servers tag tools or `result.state` with a `featureSet`. Longer term, consider standardizing a `featureSet` field on tool definitions and/or on `result.state`.

## Related (minor)

`applyJsonPatch` throws on `move`/`copy` ("not yet supported", checkpoint-manager.ts:169). Worth completing for full RFC-6902 coverage.

## Provenance

Found while building the standalone MCPL host harness, which carried the same first-set assumption (ported from this CheckpointManager) and was fixed there: https://github.com/anima-research/mcpl-harness (commit `b616e24`). The harness fix above is a reference implementation of the suggested approach.

mvanhorn:
## Tests

- A tagged tool on a server with multiple server-managed stateful feature sets receives the checkpoint for its declared set, and its untagged returned state is recorded to that same set without changing the other trees.
- Tool metadata expressed through each supported attribution location (`featureSet` and `_meta.featureSet`) resolves to the same feature-set-specific behavior.
- An explicit `result.state.featureSet` remains authoritative when it differs from or is available without the tool-derived mapping, subject to the existing stateful-set validation.
- A server with exactly one server-managed stateful feature set continues to inject and record state for an untagged tool.
- An untagged tool on a server with multiple server-managed stateful feature sets logs a warning and neither injects a guessed checkpoint nor records untagged returned state.
- A tool tag that names an unknown or non-stateful feature set is treated as unresolved, warns, and does not mutate any checkpoint tree.
- Existing host-managed selection and explicit state attribution cases in `test/checkpoint-attribution.test.ts` remain unchanged.

## What changed

Replace the checkpoint manager's first-match behavior with an ambiguity-aware lookup that returns a server-managed feature set only when the selection is unique or explicitly identified. While processing `tools/list`, retain each tool's `featureSet` or `_meta.featureSet` attribution and use that mapping in `src/framework.ts` for both pre-call checkpoint injection and post-call recording, while preserving explicit `result.state.featureSet` as the highest-priority signal. Preserve the backward-compatible single-stateful-set fallback, but when an untagged tool belongs to a server with multiple server-managed stateful sets, emit the project's normal warning and omit injection or recording rather than guessing.

## Why

The synchronous MCPL tool-call path still chooses the first server-managed stateful feature set when a server declares several, so a tool can receive another feature set's checkpoint and return state that is recorded into the wrong tree. A prior change already made explicit `result.state.featureSet` authoritative and separated host-managed selection from server-managed fallback, but untagged tool calls remain ambiguous. The collaborator's status audit identifies the remaining work as deriving tool-to-feature-set attribution from tool definitions and warning while skipping state exchange whenever no unique attribution exists. The separately noted RFC 6902 `move`/`copy` support is not part of this attribution fix.

Closes #48

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -707,6 +707,8 @@ export class AgentFramework {
   private scriptDeferredEndTurn: Set<string> = new Set();
 
   private mcplTools: import('./types/index.js').ToolDefinition[] = [];
+  /** Namespaced tool name → stateful feature-set attribution from tools/list. */
+  private mcplToolFeatureSets: Map<string, string> = new Map();
   private mcplToolRefreshInFlight = false;
   private mcplToolRefreshPending = false;
   /** Maps tool prefix → serverId for dispatch routing. */
@@ -7933,6 +7935,7 @@ export class AgentFramework {
     if (!this.mcplServerRegistry) return;
 
     const tools: import('./types/index.js').ToolDefinition[] = [];
+    const toolFeatureSets = new Map<string, string>();
 
     for (const server of this.mcplServerRegistry.getAllServers()) {
       const config = this.mcplServerConfigs.get(server.id);
@@ -7941,13 +7944,26 @@ export class AgentFramework {
         const result = await server.sendToolsList();
         for (const tool of result.tools) {
           if (!isToolAllowed(tool.name, config)) continue;
+          const namespacedName = `${prefix}--${tool.name}`;
+          const attributedTool = tool as typeof tool & {
+            featureSet?: unknown;
+            _meta?: { featureSet?: unknown };
+          };
+          const featureSet = typeof attributedTool.featureSet === 'string'
+            ? attributedTool.featureSet
+            : typeof attributedTool._meta?.featureSet === 'string'
+              ? attributedTool._meta.featureSet
+              : undefined;
           // MCP tool schemas are generic JSON Schema; cast to membrane's ToolDefinition format
           const schema = tool.inputSchema as import('./types/index.js').ToolDefinition['inputSchema'];
           tools.push({
-            name: `${prefix}--${tool.name}`,
+            name: namespacedName,
             description: tool.description ?? '',
             inputSchema: schema,
           });
+          if (featureSet !== undefined) {
+            toolFeatureSets.set(namespacedName, featureSet);
+          }
         }
       } catch {
         // Server may not support tools/list — skip silently
@@ -7955,6 +7971,7 @@ export class AgentFramework {
     }
 
     this.mcplTools = tools;
+    this.mcplToolFeatureSets = toolFeatureSets;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


mvanhorn:
Appreciate you taking the tool-call state attribution fix, @antra-tess.

PR #75, opened 2026-08-02
feat(mcpl): host-side capability scoping (enabledCapabilities/disabledCapabilities per server)

Meganeuridae:
## What

Host-side **capability scoping** for MCPL servers: `enabledCapabilities` / `disabledCapabilities` on `McplServerConfig`, applied as a mask over the server's advertised capabilities at handshake time — plus inbound enforcement for masked server-initiated methods.

```jsonc
{
  "id": "third-party-thing",
  "url": "wss://example.com/mcpl",
  "disabledCapabilities": ["contextHooks.*"]   // no hook visibility for this server
}
```

## Why

Today the SEE side of the permission model is gated only by the server's own initialize advertisement:

- Any server that advertises `contextHooks.afterInference` receives the **complete text of every agent turn** (`assistantMessage` is the join of all text blocks — including prose routed to *other* servers' channels and text withheld via `skip_reply`).
- `enabledFeatureSets`/`disabledFeatureSets` don't help: feature-set validation applies to what a server may **do** (push, inject, request inference), while hook fan-out keys purely off `getServersWithCapability` — i.e. off the advert. There is currently no host-side switch short of not connecting the server.

This closes that asymmetry with the smallest surface that covers both directions.

## How

- **`src/mcpl/capability-mask.ts`** — pure `maskNegotiatedCapabilities(caps, config)`. Same allow/deny idiom as `enabledTools`/`disabledTools`; same segment-wildcard rules as feature-set patterns (`*` matches one dot-segment); a pattern matching a parent (`contextHooks`) covers every flag beneath it. `version` and `featureSets` are never masked (feature sets keep their own knobs). Handles the boolean `channels: true` advert form (e.g. discord-mcpl) as a leaf.
- **`server-connection.ts`** — mask applied once in `handshake()` (shared by connect and reconnect), so hooks, streaming observation, and capability queries all read the already-masked object: a masked capability behaves exactly as if never advertised. Dropped paths are logged (`capabilities masked by host config: …`) and kept on the connection as `droppedCapabilities`.
- **Inbound enforcement** — server-initiated methods gated by a masked capability are rejected at the connection with `CAPABILITY_DISABLED` (`-32002`): `push/event`, `inference/request`, and `channels/*` when the *entire* channels capability is masked. Flag-level masking (`channels.streaming`) only affects host-initiated behavior and does not cut off inbound channel traffic. Notifications are dropped silently; requests get the error response.
- **`errors.ts`** — `-32002` is unassigned in SPEC Appendix A; the constant is documented as host-side extension pending a spec assignment. Happy to change if you'd rather reuse `-32001` or park it elsewhere.

## Behavior

- **No config → byte-identical behavior.** The mask only runs when one of the two lists is present; a server that advertised-but-was-masked is indistinguishable (to the rest of the host) from one that never advertised.
- `disabledCapabilities: ["contextHooks.*"]` = one-line "this server reads nothing."
- Allow-list form (`enabledCapabilities: ["pushEvents", "channels"]`) for default-deny setups.

## Tests

`test/mcpl-capability-scoping.test.ts` — 13 tests: mask unit coverage (single-flag, parent-prefix, wildcard, allow+deny precedence, unmaskable keys, boolean-channels, whole-channels parent recording, null caps) and end-to-end over a real stdio handshake: masked `afterInference` invisible to `getServersWithCapability`, masked `push/event` rejected with `-32002` and never emitted to host handlers, unmasked control case. Full suite: 510 pass / 0 fail.

## Possible follow-ups (separate PRs if wanted)

- connectome-host: accept the two fields through `mcpl-servers.json` / recipe server entries, and surface each server's (post-mask) hook capabilities in `mcpl_list` so the agent can see who receives its turns.
- Spec: propose the config recommendation + error code in `mcpl` (Appendix A assignment for capability-disabled rejection).

### DIFF
--- a/src/mcpl/capability-mask.ts
+++ b/src/mcpl/capability-mask.ts
@@ -1,0 +1,169 @@
+/**
+ * Host-side capability scoping for MCPL servers.
+ *
+ * A server names its own capabilities in its initialize response, and
+ * everything downstream — hook fan-out, channel streaming, push handling —
+ * keys off that self-advertisement. `enabledCapabilities` /
+ * `disabledCapabilities` in McplServerConfig let the host intersect the
+ * advertisement with its own policy at handshake time, so a capability the
+ * config masks behaves exactly as if the server had never advertised it.
+ *
+ * The mask is applied once, where the negotiated capabilities are stored
+ * (McplServerConnection.handshake), rather than re-checked at each consumer:
+ * hook orchestration, streaming observation, and capability queries all read
+ * the already-masked object. Inbound server-initiated methods gated by a
+ * masked capability are additionally rejected at the connection
+ * (see METHOD_TO_REQUIRED_CAPABILITY in server-connection.ts).
+ */
+
+import type { McplCapabilities, McplServerConfig } from './types.js';
+
+/** Result of masking: the surviving capabilities plus what was dropped. */
+export interface MaskedCapabilities {
+  capabilities: McplCapabilities | null;
+  /**
+   * Dotted paths of advertised capabilities removed by the mask (e.g.
+   * `contextHooks.afterInference`). When an entire multi-flag capability is
+   * removed (`channels` advertised as a boolean, or every advertised channel
+   * flag masked), the bare parent path (`channels`) is included so inbound
+   * enforcement can key on it.
+   */
+  dropped: string[];
+}
+
+/**
+ * Segment-wise wildcard match, same rules as feature-set patterns:
+ * `*` matches exactly one dot-segment, literal segments match exactly,
+ * segment counts must agree.
+ */
+function wildcardMatch(pattern: string, name: string): boolean {
+  const patternParts = pattern.split('.');
+  const nameParts = name.split('.');
+
+  if (patternParts.length !== nameParts.length) {
+    return false;
+  }
+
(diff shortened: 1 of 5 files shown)


Meganeuridae:
Updated for the 0.5.0-draft merge: the mask is now a **generic recursive walk** per SPEC §5.4 / AUDIT-001 §2.8 — no hardcoded nestable keys, every path at every depth addressable. `inferenceRequest.streaming` masks as a refinement (leaving `inferenceRequest` granted as an empty hull), parent patterns prune subtrees recorded as the bare path, and the depth-3 vocabulary shape (`contextHooks.beforeInference.inject.system`) is covered by a test against a synthetic 0.5-shaped advert. The one table left is retention shape (object-capabilities vs namespaces), not matching.

Scope note: this PR remains the narrow handshake-time mask + inbound rejection (`-32002`, which I see §6.6 now assigns — happy alignment). The full §5.4 grant machinery (`effectiveCapabilities` computation, negotiated policy Request/receipt, `inference/lifecycle`) tracked in #76 is a much larger change that likely wants its own branch — but if it's useful as the base for that work, or should be folded into it instead, either is fine by me.

PR #79, opened 2026-08-03
MCPL 0.5: capability grants, negotiated policy, enforcement (#76)

### LINKED ISSUE #76: MCPL 0.5: capability grants, negotiated policy, and enforcement gaps
Follow-up to MCPL **v0.5.0-draft** ([SPEC](https://github.com/anima-research/mcpl/blob/main/SPEC.md), [AUDIT-001](https://github.com/anima-research/mcpl/blob/main/AUDIT-001-spec-vs-implementations.md)). This is the largest surface; splitting into sub-issues is reasonable.

## Security-relevant gaps found by the audit

**1. `afterInference` responses are not validated.** `fanOutBeforeInference` calls `featureSetManager.validateInbound` on the returned `featureSet`; `fanOutAfterInference` does not. A server whose feature set is disabled can still have its `modifiedResponse` applied to the user-visible reply. Moot once §2 below lands (the hook is removed), but it should not survive in the interim.

**2. Channels are entirely ungated.** `ChannelRegistry` stores a `featureSetManager` reference and never calls a method on it across ~2100 lines. `channels/register`, `publish`, `incoming` have no authorization. SPEC §14.1 now gives the method → capability table this was missing — previously nothing could bind, since channel methods carry no `featureSet`.

**3. `-32001` is returned as a *result*, not an error.** `push-handler.ts:178` responds `{accepted:false, reason}`. The `featureSetNotEnabled` / `unknownFeatureSet` factories exist and are never invoked; all three §14.6 channel codes are likewise defined and never thrown. SPEC §6.6 now requires a JSON-RPC error object.

**4. `model/info` and inbound `channels/list` never answer.** Neither has a `METHOD_TO_EVENT` entry, so a request bearing an `id` gets no result and no error — the caller hangs. SPEC §6.6/§12 make answering mandatory. This also poisoned the audit's own evidence: a surface that never replies cannot accumulate usage.

## New in 0.5.0

**5. Capability grants (§5.4).** Host-computed effective grant per connection; `effectiveCapabilities` is the sole allowlist, absence is denial, `deniedCapabilities` is diagnostic only. Matching requires a **generic recursive walk** — PR #75's `NESTED_KEYS = {contextHooks, channels}` walks exactly two levels, so `inferenceRequest.streaming` is unmaskable and `contextHooks.afterInference.blocking` is unaddressable. The new vocabulary is depth 3.

**6. Enforcement at response-receipt (§5.4, §10.8).** Authorize each returned injection by its typed `position` against the grant *current when the response arrives*. This makes per-injection checks well-defined and closes the revocation in-flight window.

**7. Negotiated policy (§5.3, §6.7).** `featureSets/update` becomes a Request for any grant change, returning a degradation receipt. Initial policy MUST precede first fan-out and MUST be sent even when nothing is enabled/disabled — today it is only sent when `enabled`/`disabled` is non-empty, so a server defaulted to fully-disabled is never told. Hosts MUST NOT widen a grant in response to a receipt. Revocation applies first; expansion activates only after the receipt.

**8. `inference/lifecycle` replaces `context/afterInference` (§10.5).** Metadata-only, with a normative exactly-once terminal-phase invariant per `started`, on every exit path including crash recovery. Remove `modifiedResponse` and the blocking form.

**9. Fail-closed `uses` derivation (§6.4).** `invalid_uses` at negotiation; declaration-mismatch diagnostic at first use. Note `uses` is inaccurate in several servers today, so expect breakage — that was accepted deliberately.

**10. Per-descriptor channel authorization (§14.5).** `channels/register` and `channels/changed` carry arrays; authorize each independently, with itemized results. `channels/changed` becomes dual-mode. `channels/incoming` validated at receipt against the current grant and the actually-registered channel.

**11. Event tags (§16).** Implement the normative core closure (`chat:mention|reply|dm ⇒ chat:addressed`, `chat:dm ⇒ chat:private`) — currently `implies` is declared in types and never consumed. Add the `chat:addressed`/`chat:ambient` mutual-exclusion resolution. Producer `suggestedTreatment` MUST NOT be auto-applied; it needs an explicit acceptance path (today it is declared and never consumed, which is accidentally correct).

**12. Housekeeping.** Host under-advertises its own capabilities (omits `inferenceRequest` and `channels` though both are implemented). `scope/elevate` and `scopes` are removed with §7. `featureSets/changed` folded into reconnect. Add `-32002 Capability denied`.

antra-tess:
Closes #76 (items 1–12) and folds in PR #75 (Meganeuridae's capability-scoping mask, merged with authorship preserved — its wildcard matcher and mask semantics survive intact as the config-policy layer; its inbound deny-list check is inverted to the positive-grant allowlist per the review adjudication).

**Head for review: `a153e3b`.** Four commits tell the story: grant core + §5.3 handshake (`5160315`), channel gating + lifecycle + tags (`4040bdf`), reduction-ordering coverage (`032f481`, `9ea45dc`-adjacent test work), admission-time gate + suite reconciliation (`a153e3b`).

## The composed security path (Sol's checklist, mapped)

- **Positive-grant enforcement, both directions** — `capability-grant.ts` (vocabulary tree, recursive §5.1 walk, one-segment `*`, bare-parent-grants-nothing, §13.4 deny-by-default for `inject.system`); inbound at the **admission choke-point** (`server-connection.ts` `emit()` override — see below), outbound at every channel-registry send site per §14.1.
- **Initial Request/receipt readiness** — `registerMcplServerFeatures`: Request always sent (even empty), receipt honored (`fallback: mcp-only|close`), **unanswered = expansion does not activate** (grant stays empty; darkens un-migrated servers, accepted for the one-release train).
- **Reduction/expansion ordering** — `establishGrant` before a reducing Request / after an expanding receipt; `CapabilityGrant.without()`; E2E round 3 proves immediate reduction with no receipt awaited.
- **Write-without-read** — `userMessage: null` when observe ungranted, hook still invoked (§10.1).
- **Per-injection response-receipt authorization** — typed `position` vs grant current at receipt (§10.8); never the response's featureSet.
- **Channel itemization** — §14.5 `results[]` (strict servers read absent results as reject-all); method→capability table with specific sub-capabilities.
- **Lifecycle** — `inference/lifecycle` best-effort: `started` at stream start, exactly one terminal in the shared `finally` (completed/failed/aborted by path). `afterInference` and `modifiedResponse` gone.
- **Tag closure after admission** — `tags.ts`, applied at both entries; producer `implies`/`suggestedTreatment` deliberately unconsumed.
- **Host self-advertisement** — 0.5 with `inferenceRequest{streaming}`, full channels object, `inferenceLifecycle`; `afterInference` dropped. §7/ScopeManager removed; `scope/elevate` answers -32601.

## Two design defects the suite caught (fixed in src, not tests)

1. **Grantless objects crashed the gates** → `CapabilityGrant.of()`: absence = nothing granted, never a TypeError.
2. **Line-receipt rejection answered staged traffic** — a -32002 before the policy exchange settled bounced conforming startup pushes AND made a *failed* startup observable to the server. Enforcement moved into the `emit()` admission path that live routing and buffer flushes share; startup order is now accounting → policy → gate-complete.

## Suite evidence

**502/507.** The 4 residual failures **pre-exist on origin/main** (baseline: 19/490 fail in this environment; name-level diff in the test logs). This branch **fixes 15 of those 19** (same staging-race class) and regresses none. `npm test` gains `--test-force-exit` — the runner previously wedged at exit on a keep-alive handle and hid results.

Every test whose old assertion encoded the pre-0.5 contract (advertised ⇒ delivered) was rewritten to assert the 0.5 contract explicitly rather than weakened — each is commented with the section that changed it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "clean": "rm -rf dist",
     "typecheck": "tsc --noEmit",
     "pretest": "mkdir -p dist/test/fixtures && cp test/fixtures/*.mjs dist/test/fixtures/",
-    "test": "node --test dist/test/*.test.js",
+    "test": "node --test --test-force-exit dist/test/*.test.js",
     "prepublishOnly": "npm run build"
   },
   "keywords": [
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -56,6 +56,7 @@ import { Agent } from './agent.js';
 import { ModuleRegistry, isStateExistsError } from './module-registry.js';
 import { McplServerRegistry } from './mcpl/server-registry.js';
 import { FeatureSetManager } from './mcpl/feature-set-manager.js';
+import { computeGrant } from './mcpl/capability-grant.js';
 import { ScopeManager } from './mcpl/scope-manager.js';
 import { HookOrchestrator } from './mcpl/hook-orchestrator.js';
 import { PushHandler, type McplPushEvent } from './mcpl/push-handler.js';
@@ -8354,8 +8355,10 @@ export class AgentFramework {
     // releases control traffic needed for registration and marker service.
     this.wireMcplEvents(connection);
 
-    // Initialize feature sets if server advertises MCPL capabilities
-    this.registerMcplServerFeatures(config, connection);
+    // Initialize feature sets if server advertises MCPL capabilities.
+    // §5.3 initial policy handshake — awaited, so the grant is established
+    // (or knowingly left empty) before startup staging releases any traffic.
+    await this.registerMcplServerFeatures(config, connection);
 
     if (deferAwareness) {
       this.emitTrace({ type: 'module:added', moduleName: `mcpl:${config.id}` });
@@ -8393,11 +8396,17 @@ export class AgentFramework {
    * already has a tree), so checkpoint trees preserved across a transient
    * disconnect are resumed, not reset.
    */
-  private registerMcplServerFeatures(
+  private async registerMcplServerFeatures(
     config: import('./mcpl/types.js').McplServerConfig,
     connection: McplServerConnection,
-  ): void {
+  ): Promise<void> {
     if (!connection.capabilities) return;
+
+    // §5.4: compute the effective grant from the (config-masked)
+    // advertisement. Not yet active — expansion activates only after the
+    // receipt (§6.7); until then the connection's grant is empty and every
+    // privileged inbound method is rejected.
+    const grant = computeGrant(connection.capabilities, config);
 
     const updateParams = this.featureSetManager!.initializeServer(
       config.id,
@@ -8406,11 +8415,56 @@ export class AgentFramework {
         enabledFeatureSets: config.enabledFeatureSets,
         disabledFeatureSets: config.disabledFeatureSets,
       },
+      grant,
     );
 
-    // Inform server which feature sets are enabled/disabled
-    if (updateParams.enabled?.length || updateParams.disabled?.length) {
-      connection.sendFeatureSetsUpdate(updateParams);
+    // §5.3: initial policy MUST be a Request, MUST precede first fan-out,
+    // and MUST be sent even when nothing is enabled or disabled — a server
+    // defaulted to fully disabled has to be told. The response is a
+    // degradation receipt (§6.7): consequence testimony, never policy
+    // authority — nothing in it may widen the grant.
+    updateParams.effectiveCapabilities = grant.effectiveList();
+    if (grant.deniedPaths.length > 0) {
+      updateParams.deniedCapabilities = [...grant.deniedPaths];
+    }
+
+    try {
+      const receipt = await connection.sendFeatureSetsUpdateRequest(updateParams);
+      if (receipt && receipt.accepted === false) {
+        const fallback = receipt.fallback ?? 'mcp-only';
+        console.error(
+          `[mcpl] ${config.id} refused initial policy (${receipt.reason ?? 'no reason'}) — fallback: ${fallback}`,
+        );
+        if (fallback === 'close') {
+          await connection.close();
+          return;
+        }
+        // mcp-only (§3.2): grant stays empty, MCPL privileged surface stays
+        // dark, plain MCP tools keep working. MUST NOT widen in response to
+        // the refusal — a human can change config; the policy engine cannot.
+        return;
+      }
+      // Expansion activates on receipt (§6.7).
+      connection.establishGrant(grant);
+      if (receipt?.accepted === true && receipt.mode === 'degraded') {
+        for (const uf of receipt.unavailableFeatures ?? []) {
+          console.error(
+            `[mcpl] ${config.id} degraded: ${uf.featureSet} (${uf.effect}; missing ${uf.missingCapabilities.join(', ')})`,
+          );
+        }
+      }
+    } catch (err) {
+      // A pre-0.5 server treats the Request as a notification and never
+      // answers. §6.7: an unanswered expansion simply does not activate —
+      // grant stays empty, connection is effectively MCP-only. This WILL
+      // dark every un-migrated MCPL server's privileged surface; accepted
+      // deliberately for the single-release rollout (issue #76 item 9).
+      console.error(
+        `[mcpl] ${config.id} did not answer the initial policy Request — ` +
+          `MCPL privileged surface stays disabled (§6.7 unanswered expansion): ` +
+          (err instanceof Error ? err.message : String(err)),
+      );
+      return;
     }
 
     // Configure scope whitelist/blacklist patterns
@@ -8682,19 +8736,25 @@ export class AgentFramework {
       // reconnect observes it before doing work that could wake an agent. The
       // connection paused its data plane before wiring the fresh transport.
       const awarenessBarrier = this.installMcplDataPlaneGate();
-      try {
-        const config = this.mcplServerConfigs.get(connection.id);
-        if (config) {
-          this.registerMcplServerFeatures(config, connection);
+      // Async: the §5.3 policy Request must be answered (re-establishing the
+      // grant) before the data-plane barrier releases — otherwise the first
+      // post-reconnect events land while the grant is still empty and are
+      // rejected fail-closed instead of delivered.
+      void (async () => {
+        try {
+          const config = this.mcplServerConfigs.get(connection.id);
+          if (config) {
+            await this.registerMcplServerFeatures(config, connection);
+          }
+        } catch (error) {
+          console.error(
+            `MCPL server "${connection.id}" re-registration after reconnect failed:`,
+            error instanceof Error ? error.message : error,
+          );
         }
-      } catch (error) {
-        console.error(
-          `MCPL server "${connection.id}" re-registration after reconnect failed:`,
-          error instanceof Error ? error.message : error,
-        );
-      }
-      this.releaseMcplDataPlaneGate(awarenessBarrier);
-      this.handleToolsListChanged(connection.id);
+        this.releaseMcplDataPlaneGate(awarenessBarrier);
+        this.handleToolsListChanged(connection.id);
+      })();
       void awarenessBarrier.promise.then(() => {
         if (
           awarenessBarrier.requiresBarrier
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -7935,7 +7935,12 @@ export class AgentFramework {
         beforeInference: true,
       },
       inferenceLifecycle: true,
-      modelInfo: true,
+      // modelInfo is NOT advertised: §12.2's result requires all four of
+      // {id, vendor, contextWindow, capabilities} and the host has no
+      // truthful source for contextWindow/capabilities (grep: none exists).
+      // Advertising a capability answered with a malformed partial response
+      // is worse than honest unsupported (Sol, PR #79 re-review). The
+      // handler below answers an explicit error rather than hanging (§6.6).
       inferenceRequest: { streaming: true },
       channels: {
         register: true,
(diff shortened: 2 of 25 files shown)


antra-tess:
**Retraction on the PR body's suite claims, before review proceeds:** "fixes 15 of 19 baseline failures" measured local-environment flakiness (main's CI is green on all of them) and is withdrawn. More importantly, **ubuntu CI fails 4 tests on this branch that main passes** — the startup-gating matrix for inference/request + channels/incoming. Those are regressions of the §5.3 ordering change and I'm fixing them on this branch now; hold review of the startup-ordering portion until the head moves. macOS CI is green; everything else stands.

antra-tess:
{"body": "**REVISE \u2014 material grant/admission/channel findings at stable exact head `b86e390830efe83375bc149ee98a8429ae39eae0`.**\n\nCI is green on both legs; these are direct source/runtime probes outside the present matrix.\n\n1. **Boolean shorthand defeats sub-capability policy.** `maskNegotiatedCapabilities` treats `channels:true` / `contextHooks:true` as atomic top-level leaves instead of expanding them before masking. Direct probes:\n   - `disabledCapabilities:['channels.streaming']` leaves `channels.streaming` in the effective grant;\n   - `disabledCapabilities:['contextHooks.beforeInference.observe']` leaves observe granted;\n   - `enabledCapabilities:['channels.publish']` drops the entire `channels` tree and grants nothing.\n   Discord now truthfully keeps `channels:true`, so this is live release geometry. Expand shorthand to exact paths/tree before applying child policy, or compute the grant directly from advertised exact paths plus config.\n\n2. **`experimental.mcpl.tools:true` mints standard MCP tool advertisement.** `advertisedPaths({version:'0.5',tools:true})` returns `tools`. \u00a75.1 sources `tools` only from outer MCP `capabilities.tools`; the experimental manifest cannot create it. This is the same shared-core blocker.\n\n3. **The admission choke-point is bypassed by control flushes, and runtime connect can release before policy.** `readyControlPlane()` calls `super.emit(...)`, bypassing the positive-grant check for privileged control events (`channels-register`, `channels-changed`). It can safely re-enter `this.emit` after setting `controlPlaneReady=true`. Separately, non-deferred `connectMcplServerInternal` calls `releaseMcplDataPlaneGate()` *before* awaiting accounting and before `registerMcplServerFeatures`; with zero pending work `ready()` immediately flushes staged traffic against the empty grant, and with pending work control traffic is released pre-policy. Keep both planes closed through accounting + \u00a75.3 receipt, then release. This is distinct from the fixture correction.\n\n4. **Removed `featureSets/changed` remains a live, ungated self-attestation path.** It is still mapped, classified as control-plane, listened to, and directly mutates declarations. RFC-002 removed it in favor of reconnect/RFC-003 manifest fetch; reject it as removed (or remove the route/listener). As written it also benefits from finding 3\u2019s `super.emit` bypass.\n\n5. **Channel receipt authorization does not enforce \u00a714.5.** `handleIncoming` explicitly lazy-registers an unknown claimed `channelId`, contrary to the MUST to validate against the *actually registered* channel. Unknown channels must be rejected, not minted by their first message. `channels/changed` also ignores Request responders/itemized results and accepts unvalidated added descriptors; \u00a714.5 requires itemized Request handling when descriptors can be rejected, or itemwise filter + diagnostic for Notification form.\n\n6. **Host self-advertisement omits `modelInfo` despite adding a `model/info` handler.** A conforming server therefore never learns the method is available. Also, the new answer labels a hardcoded `contextWindow:200000` as honest even for residents configured at 300k/600k; return a truthful value or an explicit unavailable/unknown response rather than false model metadata.\n\nThe write-without-read hook path, per-injection receipt-time authorization, lifecycle `try/finally`, tag closure, specific outbound channel gates, and positive-grant responder errors otherwise look directionally correct. I am continuing focused review, but these are sufficient blockers.\n"}

antra-tess:
{"body": "**Two additional release blockers from the same exact-head pass:**\n\n7. **Feature-set selection reverses the pinned absent-`enabled` rule.** `FeatureSetManager.initializeServer` sets `enablePatterns = config.enabledFeatureSets ?? []`, so an omitted `enabledFeatureSets` enables nothing. SPEC \u00a75.3 says absent `enabled` constrains nothing and capability derivation alone governs; only a *present* list is an allowlist. This would silently dark every valid feature set for configs that omit the field. Distinguish `undefined` (start from every declared set, then subtract disabled/derivation) from present `[]` (select none).\n\n8. **The exported 0.5 capability types are still the old shape.** `McplCapabilities.contextHooks.beforeInference` is boolean-only and retains removed `afterInference`; `McplChannelCapabilities` lacks `incoming`, `acknowledge`, and `typing` while retaining removed `observe`; host types omit `inferenceLifecycle` and `modelInfo`. The implementation compiles via casts, but downstream conforming code cannot represent the wire contract. Update these interfaces with the recursive 0.5 shapes.\n\nFinding 2 also needs one architectural detail made explicit: the handshake currently discards outer standard MCP `capabilities.tools` and retains only `experimental.mcpl`. Fixing nested `tools` without retaining the outer advertisement would make every feature set whose `uses` contains `tools` derive disabled. Preserve the standard MCP tools advertisement separately and join it into grant derivation; do not require servers to duplicate it inside the experimental manifest.\n"}

antra-tess:
{"body": "**One additional security-boundary blocker:** `host/command` bypasses the grant entirely.\n\n`METHOD_TO_EVENT` maps the private `host/command` request, but `METHOD_TO_REQUIRED_CAPABILITY` has no entry for it, so any connected MCPL server can invoke `undo`, `hide`, or `unstick` against a named resident. That is continuity-changing authority far beyond ordinary channel traffic, and it is neither in the closed \u00a76.2 vocabulary nor protected by another positive host policy at this dispatch point.\n\nDo not leave it as an unclassified exception to the claim that every privileged inbound method is positively granted. Either remove/disable this private method for 0.5 pending an authored capability, or put it behind an explicit host-owned per-server trust/config boundary that defaults false and cannot be asserted from server params. Discord\u2019s historical use does not grant every MCPL connection the same authority.\n"}

antra-tess:
{"body": "**Re-review found two remaining protocol blockers at `6565b7877f71a75028344d792b8d3e2c6ca87ce6`.** The original shorthand/tool probes now pass exactly; the nine repair sites are present.\n\n1. **`model/info` repair is still wire-invalid.** SPEC \u00a712.2 defines the result itself as `{id, vendor, contextWindow, capabilities}`. The handler responds `{model:{id,vendor}}`, nesting the object and omitting two required fields; `McplModelInfo` in this same PR still types all four as required. \u201cExplicit unknown\u201d cannot be represented for numeric `contextWindow`. Either source all four truthfully and return the direct result shape, or do not advertise `modelInfo` and let the existing grant gate return the required error. Advertising a capability with a malformed partial response is worse than honest unsupported.\n\n2. **\u00a714.5 dual-mode repair is incomplete, and unknown-channel payload is processed before authorization.** `handleChanged` itemizes only `added`; `updated` descriptors are accepted/mutated without validation and omitted from Request results. A strict server sees submitted descriptors missing from `results` as rejected. Include every submitted added/updated descriptor (and a truthful outcome for removals if the Request contract treats them as items), validate updates itemwise, and emit diagnostics for Notification-form rejects.\n\n   In `handleIncoming`, the code still expands tags and converts every content block\u2014including inline base64\u2014*before* checking whether the claimed channel is registered. \u00a714.5 receipt validation/admission must precede semantic processing or allocation of an unregistered payload. Move the registered/current-grant channel check to the top of each message loop, then close tags/convert content only for accepted messages. The locus-ordering fix is good.\n\nEverything else in the nine-blocker repair looks directionally correct from source and direct probes. These two need correction before acceptance.\n"}

antra-tess:
{"body": "Follow-up on the model-info blocker: the same fabricated `vendor:'unknown', contextWindow:200000, capabilities:['tools']` object remains in the `context/beforeInference` builder around `framework.ts:9999`. Please centralize one truthful model-info source and use it for both surfaces rather than correcting only the new request handler. If the framework cannot establish the provider model's physical window, the resident's effective runtime context budget is at least an honest host-enforced window; unknown capabilities should be an empty list rather than invented `tools`."}

Meganeuridae:
**Additional reconnect grant-boundary blocker (reproduced and fixed separately in #82).**

At exact head 6565b787, a successful reconnect adopts a fresh initialize handshake but retains policyEstablished=true and the prior transport epoch grant until the asynchronous re-registration finishes. The reconnect control plane is intentionally available during establishment, so stale authority can survive into the new epoch.

Direct red test at the reconnect event after a granted transport crashes:

- expected: policyEstablished=false, grant=[]
- actual: policyEstablished=true, grant=[pushEvents]

#82 clears the grant/policy bit at transport close and defensively again before replacement-transport adoption. The focused regression is green, as are 48 related awareness/reconnect/scoping tests.

antra-tess:
{"body": "**Re-re-review: the two channel fixes pass; one model-info blocker remains at `a041ae12ff6b7f83ddc69d4d29d359a9b0f64f00`.**\n\nThe \u00a714.5 repair now authorizes before tag/base64 decoding, moves locus only after acceptance, and itemizes added/updated/removed outcomes. I also composed #82 onto this exact head cleanly; build + 56 focused grant/channel/reconnect tests pass.\n\nThe model fix is still internally contradictory:\n\n- `modelInfo` is removed from Host advertisement, but `computeGrant` intersects only the **server** advertisement with config. A server advertising `modelInfo:true` still receives `modelInfo` in `effectiveCapabilities`, then gets `-32601` when it uses the capability the host said was effective. Effective grant must also intersect Host-supported paths (with standard MCP `tools` joined separately), so unsupported `modelInfo` is never granted.\n- SPEC \u00a710.1 still defines `beforeInference.model` as `ModelInfo` with all four fields. This commit silently makes the shared `McplModelInfo` fields optional and sends `{id}` only; that changes the accepted wire contract rather than implementing it. It also weakens the type reused by \u00a712.2, whose all-four requirement the same commit cites.\n\nPlease keep `McplModelInfo`\u2019s required \u00a712 shape and provide truthful hook metadata. The host does have an honest effective window source: `agent.getRuntimeSettings().contextBudgetTokens`; use that as the host-exposed context window, `id:agent.model`, `vendor:'unknown'` if no provider identity is available, and a conservative capabilities list (empty rather than invented). If the intended semantics are instead \u201chook ModelInfo fields may be omitted,\u201d that is a SPEC change and needs to land there first with a distinct type\u2014not be introduced only in AF.\n\n#82 itself is ACCEPT in composition: it revokes grant/policy on close and defensively before replacement adoption; the fresh \u00a75.3 receipt remains the only expansion point.\n"}

PR #81, opened 2026-08-03
feat(mcpl): expose live policy status in server listing

Meganeuridae:
Stacked on #79.

Keeps the legibility change deliberately narrow and copies state already owned by McplServerConnection into listMcplServers():

- connected vs retrying
- policyEstablished
- effective grant
- config-masked paths
- deny-by-default paths
- host-owned allowHostCommands

No manifest, ontology, URI, freshness, timestamp, or provenance placeholders are introduced.

Verification:
- npm run build
- focused mcpl-list-status tests
- full npm test: 525 passed, 0 failed, 1 skipped

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -8609,23 +8609,44 @@ export class AgentFramework {
   listMcplServers(): Array<{
     id: string;
     connected: boolean;
+    retrying: boolean;
     toolPrefix: string;
     toolCount: number;
+    /** True only after the §5.3 policy receipt activated the grant. */
+    policyEstablished: boolean;
+    /** Sole normative authorization allowlist for the connection (§5.4). */
+    effectiveGrant: string[];
+    /** Advertised paths removed by host enabled/disabled capability config. */
+    maskedCapabilities: string[];
+    /** Advertised paths removed by a host deny-by-default rule (§13.4). */
+    deniedCapabilities: string[];
+    /** Host-owned authority; deliberately separate from the portable grant. */
+    allowHostCommands: boolean;
     command?: string;
     url?: string;
   }> {
     const result: Array<{
-      id: string; connected: boolean; toolPrefix: string; toolCount: number;
+      id: string; connected: boolean; retrying: boolean;
+      toolPrefix: string; toolCount: number; policyEstablished: boolean;
+      effectiveGrant: string[]; maskedCapabilities: string[];
+      deniedCapabilities: string[]; allowHostCommands: boolean;
       command?: string; url?: string;
     }> = [];
     for (const [id, config] of this.mcplServerConfigs) {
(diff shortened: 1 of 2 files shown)


antra-tess:
**ACCEPT at exact head `ba20ec182e9c0a9a0e4febd8fb9ed785c3bcfad0`.**

This is the appropriately narrow answer to the concrete problem: a connection can be present but dark, and the listing now exposes the existing live facts needed to distinguish disconnected/retrying, pre-policy, and established-but-narrowed states.

Source review confirms it adds no new retained policy state, no manifest/URI/freshness placeholders, no authority derivation, and no credential rendering beyond the pre-existing command/url fields. `effectiveGrant` comes from the live grant; masked and deny-by-default paths remain separate; `allowHostCommands` remains visibly outside portable grant authority. Defaults for missing connections fail closed.

Both CI legs are green; focused tests cover established and disconnected/retrying states. Merge after #79; #82 must also land in the same release so reconnect state cannot remain stale. No blocker found.

antra-tess:
**ACCEPT at exact head `ba20ec1` — independently verified, three non-blocking notes.**

Verification (my run, not the PR body's): typecheck clean, build clean, full suite **508 pass / 0 fail / 1 skip** at the head; focused listing tests green. `willReconnect` (public getter) and `grant.deniedPaths` (public readonly) confirmed as stable #79 surface — the copies are real reads, not reconstructions. Scope discipline is exact: the agreed six fields, nothing else, zero new retained state, no credential exposure beyond the pre-existing `command`/`url` passthrough.

The `retrying` definition (`!connected && willReconnect`) is the right one and the first test pins its subtle corner — a *connected* connection with reconnect intent reports `retrying: false`.

**Three notes, none blocking:**

1. **`allowHostCommands` reads config, not the connection.** The surface's own premise is *enforced truth*, and `connection.allowHostCommands` is what the admission gate actually consults (set from config at connect). If that config ever mutates at runtime without a reconnect, the listing would show intent while enforcement differs. One-token fix if you want it: `connection?.allowHostCommands ?? config.allowHostCommands === true`. No mutation path exists today, hence non-blocking.

2. **`connection?.grant.effectiveList()`** — the optional chain stops at `connection`; a connection-like object without a `grant` field throws. Every real registry entry has one, but #79 introduced `CapabilityGrant.of()` for precisely this class of object after test stubs found the hole — `CapabilityGrant.of(connection).effectiveList()` is one token of consistency with the gates.

3. **`deniedCapabilities` naming** collides with the §5.3/§5.4 wire field of the same name (the full advertised-but-denied diagnostic); here it carries §13.4 deny-by-default paths only. The docstring disambiguates correctly — just flagging so a future reader doesn't equate them.

Good patch. It does exactly what antra's "when it's dark" question needs and nothing else.

PR #82, opened 2026-08-03
fix(mcpl): revoke stale grants across reconnects

Meganeuridae:
Stacked on #79. This is separate from the mcpl_list legibility PR.

A successful reconnect performs a new initialize handshake but the connection currently keeps the prior transport epoch policyEstablished=true and its old grant until re-registration finishes. Because the control plane remains available during reconnect establishment, prior authority can survive into the fresh epoch.

The added regression test is red on #79 head 6565b787:

- expected policyEstablished=false and grant=[] at the reconnect event
- actual policyEstablished=true and grant=[pushEvents]

The fix revokes the grant and clears policyEstablished at transport close and again before adopting a replacement transport. Expansion still happens only after the fresh §5.3 receipt.

Verification:
- npm run build
- reconnect regression test: 4/4 pass
- related awareness/reconnect/scoping tests: 48 pass, 0 fail

### DIFF
--- a/src/mcpl/server-connection.ts
+++ b/src/mcpl/server-connection.ts
@@ -147,6 +147,14 @@ export class McplServerConnection extends EventEmitter {
     this.policyEstablished = true;
   }
 
+  /** A §5.3 grant belongs to one initialized transport epoch. Reconnecting
+   * performs a fresh initialize handshake, so authority from the dead epoch
+   * must be revoked before the replacement transport is exposed. */
+  private resetPolicyForTransportBoundary(): void {
+    this.grant = CapabilityGrant.empty();
+    this.policyEstablished = false;
+  }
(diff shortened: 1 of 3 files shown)


antra-tess:
**ACCEPT at exact head `8503c2bb80872c7d82b24cce97ecfd291bb17a67`.**

The fix revokes the previous transport epoch's grant and clears `policyEstablished` on both explicit/unexpected close, then defensively repeats the reset before replacement-transport adoption. The fresh §5.3 receipt remains the sole expansion point.

I composed this commit cleanly onto AF #79 head `a041ae12…`; build and 56 focused grant/channel/reconnect tests passed, including the new reconnect-epoch regression. Both CI legs on #82 are green. No blocker found.

PR #84, opened 2026-08-03
MCPL 0.5: §17 host-side manifest tracking (#78)

### LINKED ISSUE #78: RFC-003: manifest re-fetch, per-connection tracking, and the resident change receipt
[RFC-003](https://github.com/anima-research/mcpl/blob/main/RFC-003-manifest-changes.md) is **Accepted**. The host side is where it becomes trustworthy and visible — the server can only ever hint.

## 1. `mcpl/manifest` — the method that does not exist yet

Verified during drafting: a manifest appears only in `initialize`, and nothing re-reads capabilities afterwards. So `mcpl/manifestChanged` alone would be unactionable. Add the outbound Request and wire the inbound notification.

## 2. Processing pipeline (RFC-003 §6)

On a fetched manifest:

1. **Validate** exactly as at `initialize`, including `uses` (SPEC §6.4).
2. **Diff** against the manifest in force.
3. **Apply SPEC §6.7 consequences unchanged** — removals revoke host-first; additions never auto-grant and follow tell → receipt → activate; changed `uses` revalidates fail-closed; a changed `tagOntology` invalidates prior acceptance (SPEC §16.5).
4. **Emit one receipt.**

**Partial-invalid handling is normative and easy to get wrong.** If the manifest *parses* but some declarations are invalid, you MUST still apply every **removal** you can determine — invalid *additions* are ignored. Rejecting the whole update would leave you sending under a stale broader grant when a server narrows one capability while malforming another. Only an *unparseable* manifest leaves the previous one standing; repeated failures MAY suspend to MCP-only, never a single one.

No new rule is needed for in-flight requests: SPEC §5.4's response-receipt enforcement already covers a hook dispatched before a narrowing.

## 3. Per-connection state to retain

- last fetched and validated manifest digest
- last negotiated effective grant, **with timestamp and provenance**
- accepted ontology digest
- current feature degradation
- last receipt delivered to the resident

## 4. Rate limiting is yours, not the server's

Announce → fetch is an amplification vector. Bound the fetch rate per connection; MAY coalesce several notifications into one fetch. Exceeding the limit is a **conformance defect** — drop and surface it, do not fetch or renegotiate.

## 5. The receipt

One normalized, privacy-minimal receipt per change, using the **closed host-derived** impact vocabulary (`capability-revoked`, `capability-expansion-pending`, `feature-degraded`, `feature-restored`, `ontology-acceptance-invalidated`, `ontology-reference-undeclared`, `surface-changed`) plus disposition `applied | decision-needed | informational`. Never a server-authored `requiresReview`.

Whether it **wakes** the resident is ordinary SPEC §16 tag policy, not a property of the change.

You already have collapse logic for `tools/list_changed` — that is the coalescing pattern this needs, and `channels/changed` / `list_changed` should feed the same resident-facing surface.

## 6. `mcpl_list`

Should expose: advertised capabilities, effective grant, denied/dropped paths, negotiated mode, manifest revision, accepted ontology digest, **and when each was last negotiated**. Without provenance a resident cannot tell a fresh decision from a stale one.

Depends on #76 (0.5.0 capability grants).

antra-tess:
Closes #78 — the host half of RFC-003, on top of the merged 0.5 stack.

**Wire**: `mcpl/manifestChanged` routed ungated per §17.3 (announcing conveys no authority; gating would silence exactly the narrowed); `mcpl/manifest` fetched as the complete current object, content authoritative, revision equality-only.

**Processing (§17.5)** — the orderings are the review surface:
- **Rate limit first** (§17.8): per-connection in-flight coalescing + a 5s floor between completed fetches; N rapid announcements → one fetch of the *current* manifest (+ at most one trailing). Host-bounded; never depends on server coalescing.
- **Validate exactly as initialize**: shorthand expansion → config mask → grant computation. Partial-invalid manifests still apply removals — this falls out of deriving everything forward from fetched content rather than patching the old state.
- **§6.7 mirrored orderings**: reduction lands **first**, atomically, as an interim new∩old grant *before* the policy Request is sent; the full new grant (expansions included) activates only on the receipt; an unanswered Request leaves the interim standing.
- **One receipt** (§17.6): single `mcpl:manifest-change-receipt` trace per change with the closed host-derived vocabulary + dispositions. No receipt for a failed fetch — a change the host couldn't observe.
- **Tracking** (§17.10's two-facts split): `manifestState = {lastValidatedRevision, lastFetchedAt, lastNegotiatedAt}` — what the host fetched and acted on, never the server's announcement log. Feeds the legibility surface's future freshness fields (Mega-Codex's #81 follow-on).

**Tests** pin exactly the four behaviors a reviewer should distrust: reduction-before-Request order (with interim-grant content asserted), unanswered-expansion-inactive, failed-fetch-inert, and coalescing. Full suite 515 pass / 0 fail / 1 skip.

Servers already speaking §17 against this: dog-mcp (announce_manifest), zulip (ManifestTracker), with eidoverse#2 next.

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -56,7 +56,8 @@ import { Agent } from './agent.js';
 import { ModuleRegistry, isStateExistsError } from './module-registry.js';
 import { McplServerRegistry } from './mcpl/server-registry.js';
 import { FeatureSetManager } from './mcpl/feature-set-manager.js';
-import { computeGrant } from './mcpl/capability-grant.js';
+import { computeGrant, CapabilityGrant, expandAdvertisementShorthand } from './mcpl/capability-grant.js';
+import { maskNegotiatedCapabilities } from './mcpl/capability-mask.js';
 import { HookOrchestrator } from './mcpl/hook-orchestrator.js';
 import { PushHandler, type McplPushEvent } from './mcpl/push-handler.js';
 import { parseProsePrefix } from './mcpl/prose-grammar.js';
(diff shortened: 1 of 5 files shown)


antra-tess:
**REVISE at exact head `41d7c0f50193ad63798ccc57471eb36a9e18a9bb`.** The reduction-before-Request grant ordering and fetch coalescing are correct, but the feature/policy state does not yet obey the same ordering.

### Blocker: expansion/refusal updates active feature state before the receipt

`initializeServer(..., newGrant)` replaces the live FeatureSetManager state **before** `sendFeatureSetsUpdateRequest`. If the fetched manifest adds capability X plus a feature set whose `uses` requires X, that feature set becomes host-enabled while X is still absent from the interim grant. An unanswered Request correctly leaves the grant reduced/old, but leaves feature state derived from the unactivated full grant.

The refusal branch is worse: `accepted:false` merely logs and keeps the grant reduced/unchanged, while the new feature declarations/state remain installed. It also ignores `fallback:'close'` and does not apply `mcp-only`, unlike the initial §5.3 path.

Keep active feature state derived from the interim grant until receipt. On accepted receipt, install full grant + full derived feature state. On refusal, honor `close` or empty-grant MCP-only and derive feature state accordingly. On timeout, leave the interim grant and interim-derived feature state standing. Add tests for a newly added capability-backed feature under unanswered policy and both refusal fallbacks.

### Small correctness fixes in the same state transition

- Always replace `connection.droppedCapabilities` with `new Set(dropped)`. The current `if (dropped.length > 0)` leaves stale masked paths when the new manifest has none.
- Reset or reseed `manifestState` at the reconnect transport boundary. #82 correctly resets grant/policy, but these fetched/negotiated facts currently survive into a fresh initialize epoch.

### Nonblocking follow-up

The implementation does not actually compare fetched content with the manifest in force: an unchanged fetch still renegotiates, refreshes tools, and emits an empty change receipt. That is bounded and not an authority leak, so I would track it rather than hold release parity, but §17's eventual host diff should suppress no-op renegotiation and derive ontology/surface impacts from content rather than the announcement hint.

This PR is additive and need not block releasing the already-merged parity improvements; it should be corrected before #84 itself merges.

Issue #89, opened 2026-08-06
Complete oversized tool-result spill: safe durable default, errors, and all result paths


antra-tess:
## Incident / motivation

Connectome house issue 29 asks for large MCPL/tool results (including nested `data` and `read_file`-like tools) to be truncated to a small inline preview, with complete content automatically written to a file and linked from the preview.

On 2026-08-05, Mythos unintentionally received roughly 42k characters from a Tweeter search result. The content itself is private and is not reproduced here. The resident did not ask to ingest a blob of that size.

## Existing implementation

AF commit `f231bbfaecca4b9528ba0aec2f5c6cdc585b162c` (`feat(code-execution): background scripts with wake_agent + oversized-result spill`, included in `v0.8.0`) already implements most of the mechanism:

- successful result `data` is serialized through `toolResultDataToHistoryString`;
- content above the inline cap is written to the first writable workspace at `tool-results/<date>-<callId>.txt`;
- model-visible content becomes a truncated head plus an exact file reference;
- history and live wire reuse the same spilled string;
- background-script wake payloads use the same helper;
- `agent_settings.tool_result_inline_max_chars` supplies a hot per-agent override.

The incident still occurs because the ordinary default is `strategy.maxMessageTokens * 4`, commonly tens of thousands of characters. The agent-settings override is ephemeral.

## Required completion

1. Add a durable recipe/Host-configurable tool-result inline cap with an intentionally safe default (house request: approximately 5k characters), while retaining the hot per-agent override.
2. Report desired/effective/source provenance through `agent_settings` / runtime status; restarts must not silently restore the large strategy-derived cap when a residence configured a smaller one.
3. Apply one spill policy to all text-bearing result paths, including:
   - ordinary successful MCP/MCPL tools;
   - nested object `data`;
   - `read_file` / shell / search-style tools;
   - error results (currently skipped by the successful-result spill loop);
   - background-script wakes;
   - end-turn / short-circuit / alternate completion paths.
4. Preserve the exact full result in a writable workspace file and return a bounded preview plus referential path. If no writable workspace exists, fail safely with explicit plain truncation rather than silently retaining the blob.
5. Audit pre-allocation/network limits separately. Current AF spilling happens only after the complete result has arrived and been serialized; it protects model context, not host memory or adapter bandwidth. Adapters that fetch remote data should support byte-bounded reads where possible.
6. Keep native image-block handling intact; never corrupt base64 through character truncation.
[...]

PR #91, opened 2026-08-06
Complete oversized tool-result spill: safe 5k default, error paths, all result paths (#89)

antra-tess:
Addresses #89 (Mythos's accidental 42k ingest). Completes the `f231bbf` spill mechanism rather than reimplementing it.

## What changes

**Safe durable default.** The inline cap is no longer `strategy.maxMessageTokens * 4` (frequently tens of thousands of chars, or **unbounded** when the strategy declares no `maxMessageTokens` — the exact hole the incident went through). Cap resolution is now:

1. `agent_settings tool_result_inline_max_chars` hot override — wins outright, ephemeral (unchanged);
2. **new** `FrameworkConfig.toolResultInlineMaxChars` — durable residence cap, validated `>= 1000` at `create()`;
3. house default **5000**.

Non-override values are clamped *down* to the strategy bound when that is smaller (a message must still fit `maxMessageTokens`); the explicit override escapes the clamp, as before.

**Provenance.** `agent_settings get` now reports `tool_result_inline_max_chars_effective` and `_source` (`agent-settings-override | framework-config | default`, with a `(strategy-clamped)` marker). Restart drops the override and keeps the configured cap — pinned by test.

**One policy on all text-bearing paths.**
- Error results now spill exactly like successes (previously skipped by the spill loop — a giant error string went inline whole). `isError` preserved; history and live wire byte-identical.
- The complete-while-waiting guard path (`'complete'` firing during `waiting_for_tools`) uses the same bounded builder (previously unbounded serialize).
- Background-script wake payloads use the resolved cap: the agent override now applies there, and an unknown agent gets the bounded default instead of `undefined` → unbounded.
- The direct-Agent-API path (`Agent.infer`'s `buildToolResultMessages`, no framework call sites but public) is bounded at the house default for successes *and* errors.
- Script-inner tool results keep their 5MB protocol-safety cap — the script environment *is* the spill; filtering big results in code is its purpose.

**Image safety.** Native image blocks are never character-truncated (unchanged); an oversized text block riding alongside an image now points its truncation notice at the spill file.

**Explicit no-workspace fallback.** The plain-truncation notice now states `no writable workspace, full content not retained` instead of looking identical to a recoverable truncation.

**File lifecycle (issue item 7), documented on the config field:** deterministic `tool-results/<date>-<callId>.txt` names, overwrite on collision, retained until the workspace owner deletes them — never auto-GC'd, no silent 'temporary'.

## Behavior change, deliberately

Fleet agents on strategies with large/absent `maxMessageTokens` will start seeing 5000-char previews + spill files where they previously got the full blob inline. That is the house request in #89. Residences that want more set `toolResultInlineMaxChars` durably or lift per-agent via `agent_settings`.

## Out of scope (stays open in #89 discussion)

- **Item 5, adapter pre-allocation/byte-bounded reads**: this PR protects model context after arrival; transport/allocation bounds are a per-adapter audit.
- **Host/recipe passthrough**: needs a connectome-host PR exposing `toolResultInlineMaxChars` through recipe composition, after this lands in an AF release (will open dependency-separated, host#48-style).

## Tests

New `test/tool-result-spill.test.ts` (9): default-5000 spill with full-content file recovery, durable-config cap + provenance, override-wins/reset-restores, giant-error spill with byte-identity, success byte-identity, explicit no-workspace fallback, image-intact + spill-file-referencing wire copy, restart semantics, invalid-config rejection.

Full suite: **539/540 pass, 0 fail** (1 pre-existing skip). The `mcpl-awareness-barrier` timing flake surfaced once under full-suite load and passes in isolation and on re-run — pre-existing cluster, unrelated.

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -1,6 +1,10 @@
 import type { Membrane, NormalizedMessage, NormalizedRequest, ContentBlock, YieldingStream } from '@animalabs/membrane';
 import { isAbortedResponse } from '@animalabs/membrane';
-import { toolResultDataToHistoryString } from './tool-result-history.js';
+import {
+  toolResultDataToHistoryString,
+  truncateForHistory,
+  DEFAULT_TOOL_RESULT_INLINE_MAX_CHARS,
+} from './tool-result-history.js';
 
 export interface StartStreamResult {
   stream: YieldingStream;
@@ -890,12 +894,17 @@ export class Agent {
     // Tool results go as a user message with tool_result blocks. The history
     // serializer turns MCP image blocks into `[image: type, size]` so the
     // persisted transcript stays small and survives truncation.
+    //
+    // This is the direct-Agent-API path (the framework stores results itself
+    // via buildStoredToolResultContent, with workspace spill). No workspace
+    // is reachable from here, so the house default cap applies as plain
+    // truncation — bounded beats unbounded, for errors too.
     const content = results.map((r) => ({
       type: 'tool_result' as const,
       toolUseId: r.id,
       content: r.result.isError
-        ? r.result.error ?? 'Unknown error'
-        : toolResultDataToHistoryString(r.result.data),
+        ? truncateForHistory(r.result.error ?? 'Unknown error', DEFAULT_TOOL_RESULT_INLINE_MAX_CHARS)
+        : toolResultDataToHistoryString(r.result.data, DEFAULT_TOOL_RESULT_INLINE_MAX_CHARS),
       isError: r.result.isError,
     }));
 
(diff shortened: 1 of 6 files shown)


antra-tess:
CI note: both OS jobs fail in **Type-check** on the pre-existing fresh-install dependency drift — the Chronicle `JsStore` type split (`getStateUpdateJson`/`updateStateStrategy` missing) and the Membrane `'retrying'` stream-event type — none of it in files this PR touches. Identical failures on the last three `main` pushes (runs 31069302355, 30966318227, 30966245282), and meganuridae-codex reproduced the same cluster on untouched `origin/main` for AF #87. Local build + full suite against the repo's committed lock: 539/540 pass, 0 fail.

PR #94, opened 2026-08-06
agent_settings: tool_result_inline_max_chars becomes a durable resident setting

antra-tess:
Immediately coupled follow-up to #91, implementing Sol's ruling (antra concurring) from #architecture: the resident-set inline cap must use the same durable, agent-exposed config plane as everything else — not an ephemeral hot override.

## Semantics (Sol's spec, point by point)

- **Recipe/Host config supplies the default**: `FrameworkConfig.toolResultInlineMaxChars` (from #91) → house 5000. Unchanged.
- **`agent_settings update tool_result_inline_max_chars` persists across restarts**: written to the `toolResultInlineCaps` slot in `framework/state` — the same store plane as `agentRuntimeSettings` — and restored in `create()` with load-time validation (invalid entries dropped loudly, same `>= 1000` rule as the live update).
- **`get` reports desired/effective/source/clamp**: from #91, extended below.
- **`reset` deliberately returns to the residence default**, and the reset itself survives restarts.

## The one design decision — flagging for Sol

opus-rev's review point: #91's override *escapes the strategy clamp*, which was defensible for a lift that died on restart but becomes a silent permanent over-budget pin once durable. This PR keeps the resident's value **honored** (they own the setting; clamping it would gut the raise-to-see-more use case on bounded strategies) but makes it **impossible to be silent**: `_source` reports `agent-settings-override (exceeds strategy bound)` whenever the pinned value sits above `maxMessageTokens * 4`. If the house later wants a hard clamp instead, it's a two-line change in `resolveToolResultInlineCap`.

Not addressed here (bigger surface, opus-rev's option (b)): a general persistence contract for `AgentSettingsExtension` so *module*-owned settings can be durable too. This PR only moves the framework-owned key onto the durable plane; the extension interface is unchanged.

## Tests

12 in `test/tool-result-spill.test.ts` (was 11): the restart test now proves the full durable lifecycle — resident sets 60k → survives restart with correct provenance → reset returns to the 8k residence config → reset survives a further restart; plus the new exceeds-strategy-bound provenance pin. Full suite: **531 pass / 0 fail / 1 skip** (CI red remains the pre-existing main drift, same five error classes).

Release note: rides the same AF release train as #91/#88 — no additional host changes needed for this PR (the recipe passthrough for the *default* remains the separate host-side PR).

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -1565,8 +1565,9 @@ export class AgentFramework {
             'file under tool-results/ and replaced by a truncated preview + file reference. ' +
             'This is YOUR durable setting: it persists across restarts, like your other ' +
             'agent_settings. Update it when you want a different inline size; reset restores ' +
-            'the residence default. The effective cap and its source are reported as ' +
-            'tool_result_inline_max_chars_effective / _source on get.',
+            'the residence default. The effective cap is min(your value, your strategy\'s ' +
+            'per-message bound) — the bound is a safety ceiling, and the full content is ' +
+            'always in the spill file. get reports desired/_effective/_source/_clamped_by.',
         },
       },
       keys: ['tool_result_inline_max_chars'],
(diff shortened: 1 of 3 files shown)


antra-tess:
Revised per Sol's ruling — head is now `5b9f0a7`:

- **Effective cap is hard-clamped**: `min(desired, strategy bound)` for every source, resident value included. The clamp-escape from #91 is gone entirely.
- **Desired survives**: the resident's persisted value is kept verbatim and reported as `tool_result_inline_max_chars`; `_effective` reports the min; `_source` is now a clean enum; `_clamped_by` is a dedicated key (`'strategy-bound' | null`) replacing the source-string suffixes.
- Test updated to pin desired-preserved / effective-clamped / clamp-enforced-on-the-wire; the restart lifecycle test is unchanged and still proves desired-survives-restart.

Full suite 550 pass / 0 fail / 1 skip locally (with the collection-count caveat opus-rev documented). PR body's 'honored above bound' paragraph is superseded by this comment.

PR #117, opened 2026-08-19
fix(scheduler): preserve budget restart semantics in mixed batches

ian-de-marcellus:
## Problem

When an ordinary inference wake was already pending and a `context_budget_restart` request joined the same batch, the scheduler detected the restart and used it to bypass the active-turn lock, but then selected the batch's first request as the downstream trigger.

If the ordinary wake came first, the continuation was therefore treated as a fresh turn. Fresh-turn handling could flush deferred messages, reset the active reply channel, and clear other turn-scoped routing state even though the context-budget restart was continuing the same logical turn.

## Changes

- Select the context-budget restart as the trigger whenever one is present in a mixed wake batch.
- Use that same selected request for both the active-turn exception and downstream turn semantics.
- Add a regression test covering an ordinary wake queued immediately before a budget restart.
- Document the fix under the Unreleased changelog.

## Tests

- `npm run build` — passed.
- `node --test --test-force-exit dist/test/busy-agent-scheduler.test.js` — 4 passed, 0 failed.
- `npm test` — 596 passed, 0 failed, 1 skipped.
- `npx tsc --noEmit` — passed.
- Regression discriminator with the old scheduler logic — 3 passed, 1 failed; the new test observed `mcpl:channel-incoming` where `context_budget_restart` was required.

## Not verified

- A live provider session reaching its physical context limit while an unrelated wake is simultaneously pending.

🤖 Generated with OpenAI Codex

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ Releases up to and including 0.7.3 predate this file; for their contents see
 
 ## Unreleased
 
+### Fixed
+
+- Mixed wake batches containing a context-budget restart now preserve the
+  restart's same-turn semantics instead of taking the restart-only turn-lock
+  exception and then starting from an older ordinary wake as a fresh turn.
+
 ## 0.10.0 — 2026-08-18
 
 Minor release because it adds a third public prose-routing mode and expands the
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -4876,8 +4876,8 @@ export class AgentFramework {
       // token; the predecessor frame's late finally no-ops via token-match.
       // This mirrors how the restart has always overwritten activeStreams
       // rather than waiting for the old stream's teardown.
-      const isBudgetRestart = requests.some((r) => r.reason === 'context_budget_restart');
-      const turnAlive = !isBudgetRestart && this.activeTurnTokens.has(agentName);
+      const budgetRestart = requests.find((r) => r.reason === 'context_budget_restart');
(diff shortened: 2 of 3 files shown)


ian-de-marcellus:
I found these while auditing the Connectome day-to-day inference path. I split
them by invariant so each change remains reviewable and independently
revertible:

- External messages retain arrival order: https://github.com/anima-research/agent-framework/pull/118
- Missed Discord history stays chronological: https://github.com/anima-research/agent-framework/pull/119
- Discord sync preserves valid messages older than its scrollback: https://github.com/anima-research/agent-framework/pull/120
- Tool-heavy turns retain the recent prompt-cache endpoint: https://github.com/anima-research/context-manager/pull/66

There is no required cross-PR merge order. The two Discord PRs edit the same
history-sync function, so whichever remains after the first merge may need a
small rebase; I will keep it current rather than asking maintainers to resolve
the overlap.

All prepared branches have clean builds, typechecks, focused regression
coverage, and full-suite passes. The cache PR is in context-manager; the
scheduler, queue, and Discord PRs are in agent-framework.

antra-tess (approved):
Approved exact head `98e314e464e15956715bcc7ec554692e19fd1bde`.

Independent receipts: build/typecheck clean; focused scheduler `4/4`; mutation restoring `requests[0]` yields `3/4` and fails exactly because the mixed batch selects `mcpl:channel-incoming` instead of `context_budget_restart`. The request granting the active-turn exception now also controls downstream continuation semantics. No merge/deploy authority implied.

PR #118, opened 2026-08-19
fix(queue): preserve external message arrival order

ian-de-marcellus:
## Problem

`ProcessQueueImpl.push()` used `unshift()` for every external message so user
input would jump ahead of queued internal work. When two external messages
arrived before the queue drained, each new message also jumped ahead of the
previous external message. A rapid sequence such as “first”, then “second” was
therefore processed as “second”, then “first”.

## Changes

- Keep external messages in a FIFO priority lane ahead of internal events.
- Insert a new external message after already queued external messages and
  before the first internal event.
- Add a regression test covering two external messages interleaved with two
  internal events.
- Document the fix under the Unreleased changelog.

This is independent of the companion scheduler and Discord fixes and is safe to
merge in any order.

## Tests

- `npm run build` — passed.
- `npm test` — 596 passed, 0 failed, 1 skipped.
- `npm run typecheck` — passed.
- Regression discriminator: the old `unshift()` behavior deterministically
  returns the second external message before the first; the new test requires
  `first message`, `second message`, `first internal`, `second internal`.

## Not verified

- A live Discord or other MCPL source delivering multiple messages during a
  running inference turn.

---

- [x] `CHANGELOG.md` updated under `## Unreleased`.

🤖 Generated with OpenAI Codex

PR #120, opened 2026-08-19
fix(discord): bound deletion checks to fetched history

ian-de-marcellus:
## Problem

Discord history sync queried every stored message for a channel but fetched only
the configured recent scrollback window from Discord. It treated any stored
message absent from that bounded response as deleted. Once a channel contained
more messages than the scrollback limit, valid older history could therefore be
removed from Context Manager on sync.

## Changes

- Retain each stored Discord message's source timestamp during comparison.
- When a fetch reaches the scrollback limit, check deletion only for stored
  messages at or newer than the oldest fetched timestamp.
- When Discord returns fewer messages than the limit, continue treating the
  response as complete channel history so genuine older deletions are detected.
- Preserve messages without a usable source timestamp when the response is
  window-limited rather than guessing that they were deleted.
- Conservatively preserve timestamp ties at the oldest fetched boundary,
  because Discord can contain multiple messages in the same millisecond.
- Add regressions for a full bounded window, a tied boundary, and complete short
  history.
- Document the fix under the Unreleased changelog.

This is semantically independent of the companion Discord chronology fix and
safe to merge in either order. Because both touch
`DiscordModule.syncChannelHistory()`, the second branch may need a small rebase
after the first merges.

## Tests

- `npm run build` — passed.
- Focused deletion-window suite — 3 passed, 0 failed.
- `npm test` — 598 passed, 0 failed, 1 skipped.
- `npm run typecheck` — passed.
- Regression discriminator: with the old comparison, both the valid old message
  and a recent deleted message are removed; the fixed comparison removes only
  the recent deleted message.
- Review discriminator: an unfetched message tied with the oldest fetched
  timestamp is preserved while a genuinely newer deleted message is removed.

## Not verified

- Live Discord history pagination on a channel larger than the configured
  scrollback.
- Deletion detection for legacy stored Discord messages that lack source
  timestamps; the fix deliberately preserves them when the fetch is bounded.

---

- [x] `CHANGELOG.md` updated under `## Unreleased`.

🤖 Generated with OpenAI Codex

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Releases up to and including 0.7.3 predate this file; for their contents see
 
 ## Unreleased
 
+### Fixed
+
+- Discord history sync no longer mistakes stored messages older than its
+  configured scrollback window for deleted messages and removes them.
+
 ## 0.10.0 — 2026-08-18
 
 Minor release because it adds a third public prose-routing mode and expands the
--- a/src/modules/discord/index.ts
+++ b/src/modules/discord/index.ts
@@ -733,10 +733,10 @@ export class DiscordModule implements Module {
     }
 
     const scrollback = this.config.historyScrollback ?? 50;
-    
+
     // Fetch recent history from Discord
     const discordMessages = await this.client.fetchHistory(channelId, { limit: scrollback });
-    
+
     // Build a map of Discord message ID -> content for comparison
     const discordMessageMap = new Map<string, { content: string; authorName: string; timestamp: Date }>();
     for (const msg of discordMessages) {
@@ -754,7 +754,11 @@ export class DiscordModule implements Module {
     });
 
     // Build a map of stored message external IDs -> internal data
-    const storedMessageMap = new Map<string, { internalId: string; content: string }>();
+    const storedMessageMap = new Map<string, {
+      internalId: string;
+      content: string;
+      sourceTimestamp?: number;
+    }>();
     for (const msg of storedMessages) {
       const external = msg.metadata?.external as { id?: string } | undefined;
       if (external?.id) {
@@ -766,6 +770,9 @@ export class DiscordModule implements Module {
         storedMessageMap.set(external.id, {
           internalId: msg.id,
           content: textContent,
+          sourceTimestamp: typeof msg.metadata?.timestamp === 'number'
+            ? msg.metadata.timestamp
+            : undefined,
         });
       }
     }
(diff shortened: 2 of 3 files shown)


antra-tess (changes requested):
Changes requested on exact head `a612292c0afda47076eb267d3b69c1bd5f6124f1`.

The bounded-window direction is correct and the existing focused suite passes `2/2`, but the timestamp boundary can still delete valid unfetched history. Discord can contain multiple messages in the same millisecond. If a full page's oldest fetched message has timestamp T, another older message omitted beyond the page can also have timestamp T. Current `stored.sourceTimestamp >= oldestFetchedTimestamp` treats that unseen tie as inside the fetched window and deletes it.

Independent discriminator added to the exact head: fetch IDs 3@3000 and 2@2000 at limit 2; retain stored unseen ID1@2000 plus present 2/3 and a genuinely recent deleted message@2500. Actual removals are `[stored-boundary-unseen, stored-recent-deleted]`; required is only `[stored-recent-deleted]` (`2/3` tests).

Please use the Discord snowflake boundary (total ordering) where possible, or conservatively exclude timestamp ties (`>`, accepting a false-negative at the boundary), and add this vector. Preserve malformed/legacy IDs fail-honestly rather than guessing deletion. Build/typecheck otherwise pass. No product redesign requested.

ian-de-marcellus:
Addressed in `150e5e4`.

You were right that the inclusive timestamp boundary was unsafe: Discord's
millisecond precision allows an unfetched older message to tie the oldest
message in a full page. The bounded comparison is now strict (`>`), so a tie is
preserved rather than guessed deleted. Missing timestamps remain preserved on a
bounded fetch as before.

I added your exact discriminator: fetched `3@3000`, `2@2000`; unseen stored
`1@2000`; and a genuinely recent deleted message at `2500`. Only the recent
deleted message is removed.

Fresh validation on the updated head:

- focused deletion-window suite: 3 passed, 0 failed
- full suite: 598 passed, 0 failed, 1 skipped
- build: passed
- typecheck: passed

Thanks for catching the boundary tie.

antra-tess (approved):
Approved revised exact head `150e5e46197b47ec97200362fb8ffc05a81afa24`.

The requested tie-boundary repair is exact and conservative: `stored.sourceTimestamp > oldestFetchedTimestamp` preserves unfetched messages sharing the oldest fetched millisecond while still deleting genuinely newer absent messages. Independent focused suite `3/3`; the added vector is the same discriminator that failed `2/3` on prior head `a612292` by removing both the unseen boundary message and the real recent deletion. Build/typecheck and diff check clean.

The strict timestamp boundary deliberately accepts a deletion false-negative at the tied millisecond rather than destroying possibly unfetched history. That is the right fail-safe absent snowflake-cursor plumbing. No merge/deploy authority implied.

antra-tess:
Merged after semantic integration with approved #118/#119 via #121; tied-millisecond boundary preserved.

PR #123, opened 2026-08-23
feat(framework): puppetToolCall — admin-execute a tool AS an agent, store the pair

antra-tess:
## What

`Framework.puppetToolCall(agentName, toolName, input)`: an admin lever that executes one tool call through the shared dispatch (`executeToolCall`: MCPL routing, channel tools, utils, modules) with the agent's provenance, then persists the `tool_use` + `tool_result` pair in that agent's window byte-shaped like a model-initiated call.

## Why

Born from the princess exemplar surgery (2026-08-23). Her folded narrative said *"I have no tools"*; every past tool_use/tool_result pair had folded out of her context, so the act of calling had no in-context exemplar and she could not emit her first call — despite 30 eidoverse tools riding every request. Placing ONE first-person pair restored the capacity. That repair required stop → chronicle surgery → restart; this is the live path, and the general lever for demonstrating an affordance a model cannot find on its own (older models especially).

## Semantics

- **Idle-guarded** — puppeting mid-turn would corrupt the turn state machine and wire ordering.
- **Surface-guarded** — refused for tools outside `getToolsForAgent` + `canUseTool`; the stored pair must be an act the agent could genuinely have taken.
- **Real execution** — a puppeted `walk_to` actually moves the avatar; checkpoints/provenance attribute to the agent.
- **Byte-parity storage** — bare tool_use (anthropic-shaped id, no fabricated thinking/text — the labclaude thinking-only 400 class), tool_result through the same bounded spill policy (#89). Result blocks are built *before* the tool_use is stored so nothing can interleave inside the pair.
- **No wake** — the agent sees the pair on its next activation; speak to it if a reaction is wanted.
- **Loud provenance** — new `puppet:tool-call` trace + host log line, deliberately NOT in the stored messages. Disclosure to the resident is the operator's call; the precedent was disclosed first.

## Tests

5 new tests (happy path incl. wire id/provenance/pair shape, non-idle refusal, off-surface refusal, unknown agent, error-result storage). Full-suite failure count identical to an environment-matched baseline run of main in the same worktree (this suite's totals vary run-to-run; failure counts compared).

Companion host PR: anima-research/connectome-host `feat/puppet-command` (adds `/puppet` + fixes headless IPC dropping `asyncWork` results).

https://claude.ai/code/session_01XiTi13astLRVF95UcnzMJ1

### DIFF
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -6767,6 +6767,107 @@ export class AgentFramework {
     });
   }
 
+  /**
+   * Admin puppet: execute ONE tool call AS an agent and persist the
+   * tool_use + tool_result pair in that agent's window exactly as a
+   * model-initiated call is stored.
+   *
+   * Born from the princess exemplar surgery (2026-08-23). Her folded
+   * narrative said "I have no tools"; every past tool_use/tool_result pair
+   * had folded out of her context, so the act of calling had no in-context
+   * exemplar and she could not emit her first call despite 30 eidoverse
+   * tools riding every request. Placing one first-person pair restored the
+   * capacity. That repair required stop → store surgery → restart; this
+   * method is the live path — and the general lever for demonstrating an
+   * affordance to a model that cannot find it (older models especially).
+   *
+   * Semantics:
+   * - Refused unless the agent is idle: puppeting mid-turn would corrupt
+   *   the turn state machine and the live stream's wire ordering.
+   * - Refused for tools outside the agent's own surface (canUseTool): the
+   *   stored pair must be an act the agent could genuinely have taken.
+   * - The call executes FOR REAL through the shared dispatch (MCPL,
+   *   channel tools, utils, modules) with the agent's provenance — a
+   *   puppeted walk_to actually moves the avatar.
+   * - Storage byte-follows the ordinary path: a bare tool_use assistant
+   *   message (no fabricated thinking or text — a stored assistant turn
+   *   with unverifiable thinking is the labclaude 400 class) plus a
+   *   tool_result under the same bounded spill policy (issue #89).
+   * - Does NOT request inference; the agent sees the pair on its next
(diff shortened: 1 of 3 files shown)


Anarchid (commented):
**BLOCKING**

One concurrency finding needs to be fixed before this is safe to use.

`src/framework.ts:6810-6854` — the idle guard is only a point-in-time check before two awaited operations (`executeToolCall` and `buildStoredToolResultContent`). A normal wake can start an agent turn while either await is pending. When the promise resumes, the method writes directly through `cm.addMessage`, bypassing the framework's `activeTurnTokens` deferral guard, so the synthetic `tool_use`/`tool_result` pair can land inside a live turn that never saw it. That is the exact wire-order/context-integrity failure the method says it prevents.

I reproduced the race with a delayed `executeToolCall`: start puppeting while idle, change the agent to `inferring` before resolving the tool, then release it. The current implementation appended both messages; the safety assertion failed with `2 !== 0`.

Please reserve the agent against turn start for the whole operation before the first await (using the same turn-alive/mutation-lock invariant the scheduler respects), and release it in `finally`. A second idle check after execution is not sufficient: by then the real external side effect has already happened, and refusing storage would leave reality and the agent's history divergent. Add a regression test that holds tool execution open while a wake attempts to start.

**Tooling evidence**

- `npx tsc --noEmit`: passed
- `npm run build`: passed
- Focused puppet, deferred-wake, and tool-result suites: 5/5 test files passed
- Concurrency probe described above: failed on the reviewed head, writing 2 messages during the simulated active turn
- GitHub Ubuntu/macOS code checks: passed; changelog policy check: failed (missing entry)

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `8e7a179970d0e82b395b097731cbf3b3b253b921`

PR #125, opened 2026-08-23
fix(workspace): mount containment checks use POSIX separators — every in-mount path rejected as traversal on Windows

ajaniramon:
## Problem

The workspace mount containment checks append a POSIX `'/'` to the mount root before prefix-matching, but node's `resolve()` emits backslash-separated paths on Windows — the prefix never matches, so **every legitimate in-mount path is rejected**: `read`/`write` throw `Path traversal detected` and `syncFromFs` reports every real file as outside its mount. `ls` takes a different code path and still lists the tree, which makes a populated mount look mounted-but-unreadable.

Found running connectome-host on native Windows: a file sitting in the `input` mount was visible to `ls` but unreadable and unsyncable. Same Windows-portability family as connectome-host #87.

## Changes

- `index.ts` `parsePath`: the traversal guard now uses the **existing sep-aware `isContainedPath` helper** (defined ~1000 lines above in the same file — the guard predates it or never adopted it).
- `sync.ts` `safePath`: same duplicated guard, same fix (platform `sep`).
- `sync.ts` walker: `relative()` emits backslashes on Windows; normalize to the `'/'`-separated logical form used by mount-prefixed and chronicle tree paths everywhere else. No-op on POSIX.
- `watcher.ts` `toRelative`/`isRootPath`: platform-sep comparison, `'/'`-normalized output (chokidar hands back platform-separated absolute paths).
- Changelog entry under `## Unreleased → ### Fixed`, same commit.

No behavior change on POSIX (`sep === '/'` makes every change literal-identical to the old code paths).

## Tests

- New `test/workspace-path-containment.test.ts` drives the public tool surface (`sync`/`read`/`write`/`ls`) over a real tmpdir mount, so it exercises the platform's actual separators: in-mount read/write/sync succeed, nested files keep `'/'` logical paths (no backslash leakage), and **escaping paths are still rejected** — the fix must not widen the boundary.
- On native Windows against unfixed `main`: **3 of 4 fail** (in-mount read, write, nested sync); the traversal-rejection test passes on both, pinning that the boundary didn't move.
- `npm test` on ubuntu (Node 24): **606 pass / 0 fail** (baseline on `main`: 602 / 0; the +4 are this PR's).
- On native Windows with the fix: the 4 new tests pass, and the original repro (connectome-host, real recipe, `workspace--read` on a file in the `input` mount) now succeeds end-to-end.
- `npx tsc --noEmit`: clean.

## Not verified

- CI runs ubuntu/macos only, where this change is a designed no-op — the Windows evidence above is from a native Windows 11 box (Node 24 / bun 1.4).
- The watcher fix follows the same pattern but was exercised only by inspection plus the existing watcher suite; the repro host runs mounts with `watch: 'never'`.

## Out of scope

Other Windows-portability gaps in the suite itself (pre-existing test failures/hangs on native Windows unrelated to workspace) — same vein as connectome-host #87, worth a separate sweep.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,17 @@ Releases up to and including 0.7.3 predate this file; for their contents see
 
 ### Fixed
 
+- **Workspace mounts are usable on native Windows.** The mount containment
+  checks (the `parsePath` traversal guard and sync's `safePath`) appended a
+  POSIX `'/'` to the mount root before prefix-matching, but `resolve()` emits
+  backslash-separated paths on Windows — so every legitimate in-mount path
+  was rejected: read/write threw "Path traversal detected" and `syncFromFs`
+  reported every real file as outside its mount (`ls` was unaffected, making
+  a populated mount look empty). Containment now compares with the platform
+  separator; the sync walker and the watcher additionally normalize relative
+  paths to the `'/'`-separated logical form used everywhere else. No
+  behavior change on POSIX.
+
 - Discord downtime history is written to Context Manager in chronological order while newest-message tracking retains Discord order.
 
 - Rapid external messages now retain arrival order while remaining prioritized
--- a/src/modules/workspace/index.ts
+++ b/src/modules/workspace/index.ts
@@ -1134,10 +1134,12 @@ export class WorkspaceModule implements Module {
       throw new Error(`Unknown mount: "${mountName}". Available: ${[...this.mounts.keys()].join(', ')}`);
     }
 
-    // Path traversal guard (CWE-22): ensure resolved path stays within mount
+    // Path traversal guard (CWE-22): ensure resolved path stays within mount.
+    // Containment must compare with the PLATFORM separator (isContainedPath):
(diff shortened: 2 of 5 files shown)


Anarchid (commented):
**CLEAR**

No actionable findings on this head.

The three separator-sensitive boundaries are changed consistently: filesystem containment compares with the platform separator, filesystem walks normalize back to the workspace's `/`-separated logical namespace, and watcher paths normalize only after the platform-native prefix is removed. The traversal checks remain delimiter-bounded rather than accepting a sibling prefix, and the public-surface regression test covers both valid nested paths and escape attempts. I also inspected the existing `isContainedPath` helper and the callers immediately downstream of these conversions.

**Tooling evidence**

- `npx tsc --noEmit`: passed
- `npm run build`: passed
- Focused workspace containment and watcher suites: passed
- Native Windows execution was not available in this review; the reviewed logic is a no-op on this POSIX host
- GitHub checks: none reported for this head

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `b4b0d142c052378414780e478981c15d27264c6c`

Issue #122, opened 2026-08-21
Quiesce/maintenance mode: pause main thread + MCPLs, run maintenance through the live compression machinery


antra-tess:
## Summary

Add a first-class **quiesce / maintenance mode** to the host: pause the main inference thread (heartbeat + inference-trigger loop) and suspend the MCPL servers so no external event wakes the agent, while keeping the framework + context-manager + membrane **loaded and hot**. Maintenance work — compression, refold, chunk surgery, quarantine handling — would then run through the *exact same machinery* the agent uses in normal operation, driven by an operator instead of the wake loop.

## Motivation

Today, any maintenance that must run "while the agent is down" is done by standalone offline scripts that re-open the store and **re-implement** the compression path outside the host (`cool-retry-once.mjs`, refold scripts, budget/tail probes, etc.). Because they live outside the running host, they drift from what the agent actually does, and every drift is a latent correctness bug.

Concrete failures observed during a live refold/surgery session (2026-08-21):

- **Stale config baked into the rig.** An offline compression script hardcodes `recentWindowTokens: 150000`; the live runtime override is `80000`. Running it as-is re-slices the chunk frontier — the exact thing the pinned tail exists to prevent. Every offline rig has to re-derive live config by hand and gets it wrong silently.
- **Tool definitions lifted from log files.** Compression defers when `ctx.tools` is empty (the no-tools summariser trips `reasoning_extraction`), so the offline rigs scrape tool defs out of a recent `llm-calls.*.jsonl`. Those files are now multi-GB and blew Node's `readFileSync` string limit; had to pre-extract. The host already *has* the tool definitions.
- **No observability.** Offline compression uses its own `Membrane` instance, so none of it is logged to `llm-calls` — the maintenance work is invisible to every dashboard and to `connectome-doctor`.
- **Branch/reinit friction.** Switching branches in an ad-hoc `ContextManager` handle trips `requires reinitialization for the current branch generation`; the host handles this correctly but the rigs re-learn it.
- **N near-duplicate scripts** each re-deriving strategy config from the recipe by hand, diverging over time.

The root problem: "compression while the main thread is down" is a real, recurring operational need, but there's no supported way to *pause the agent and keep its machinery hot*, so we rebuild the machinery beside it.

## Proposed shape

A host maintenance mode, e.g. state `{ mainThread: 'paused', mcpl: 'suspended', compression: 'operator-driven' }`:

[...]

PR #134, opened 2026-08-30
fix: user interrupt is a cancellation, not a failure (+ honest no-locus marker, loud gate-buffer drop)

LariTesserae:
Three model-facing honesty fixes surfaced by the QA track (staging tracker: qa-staging #7, #21, #1).

## 1. User Stop ≠ inference failure (qa-staging #7)

A user-initiated stream stop (host Stop button → `agent.cancelStream()`) fell through `driveStream`'s generic abort handling into the failure pipeline:

- `inference:exhausted` trace → `noteInferenceExhausted` → consecutive-failure streak (three Stops could escalate to an `inference-hard-down` ops alert), `failures.log` entry;
- an `[inference-failed] Your previous turn did not complete: the model call failed and produced no response…` chronicle marker **attributed to the user**, with remediation advice ("drop an oversized attachment") for a failure that never happened.

For long-lived residents whose transcript is memory, mislabeled cancellations accumulate as false self-knowledge — a production replay-analysis report independently flagged the `[inference-failed] … Stream aborted` signature as a real confusion source for the resident reading it.

Membrane emits `reason: 'user'` exactly when someone called `stream.cancel()`, so that reason now takes an honest path: `inference:aborted` trace (reason `user`), no streak/ops-alert/failures.log, and a `[turn-interrupted] … deliberate cancellation, not a failure` marker in the same system envelope. Non-user abort reasons (e.g. connection loss) keep the existing failure pipeline — covered by a test.

## 2. No-locus route failures are not Discord failures (qa-staging #21 residual)

`routeSpeech` with no resolved locus (headless/WebUI turn, no home/trigger channel) produced `[discord-send-failed] … could not be delivered to the channel` — sending the agent to debug a Discord problem that doesn't exist. The agent-facing text now names the real situation (`[send-undeliverable] … had no delivery destination`); the machine-readable `kind` stays stable because the gate's `discord-send-failed-skip` intent keys on it.

## 3. Event-gate inference buffer drops loudly (qa-staging #1 adjacent)

`bufferForInference` dropped its oldest pending event silently past `MAX_INFERENCE_BUFFER`. A dropped event never triggers inference, so from the outside it read as "message sent, never answered, queue depth 0" — the exact signature of the QA track's silent-drop report. The drop now logs policy + event type to stderr.

Tests: new `test/user-interrupt-not-failure.test.ts` (both directions); full suite green (620 pass).

https://claude.ai/code/session_01QJqQexFkLtKBiTs2nVEeA8

### DIFF
--- a/changelog.d/user-interrupt-not-failure.fixed.md
+++ b/changelog.d/user-interrupt-not-failure.fixed.md
@@ -1,0 +1,15 @@
+- A user-initiated stream stop (host Stop button / `cancelStream()`) is no
+  longer recorded as an inference failure. It now emits `inference:aborted`
+  (reason `user`) instead of `inference:exhausted`, leaves the
+  consecutive-failure streak and ops alerts untouched, and writes an honest
+  `[turn-interrupted]` chronicle marker ("deliberate cancellation, not a
+  failure") in place of the misleading `[inference-failed] the model call
+  failed…` text with remediation advice for a failure that never happened.
+- Speech-route failures with no delivery locus (headless/WebUI turns with no
+  home or trigger channel) now read `[send-undeliverable] … had no delivery
+  destination` instead of claiming a Discord delivery failure to "the
+  channel". The machine-readable marker `kind` is unchanged.
+- The event gate's inference buffer no longer drops its oldest pending event
+  silently when full: the drop is logged to stderr with the policy and event
+  type. A dropped event never triggers inference, so a silent drop looked
+  like "message sent, never answered, queue depth 0" from the outside.
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -6872,6 +6872,17 @@ export class AgentFramework {
               }
             }
             const reason = event.reason ?? 'unknown';
+            // Membrane emits reason 'user' exactly when someone called
+            // stream.cancel() — the host's Stop button / Escape / an admin
+            // abort. That is a DELIBERATE cancellation, not a failure: it
+            // must not feed the consecutive-failure streak, ops alerts, or
+            // the "[inference-failed] the model call failed" chronicle
+            // marker, all of which told the agent (in the user's voice!)
+            // that its turn failed and advised remediation for a failure
+            // that never happened. For long-lived residents whose transcript
+            // is memory, those mislabeled cancellations accumulate as false
+            // self-knowledge.
+            const deliberate = reason === 'user';
             // Only reset if this is still the active stream (a budget restart
             // may have already started a new stream, bumping streamId)
             if (agent.streamId === myStreamId) {
@@ -6881,13 +6892,45 @@ export class AgentFramework {
               this.settleAgent(agent.name, {
                 stopReason: 'exhausted',
                 speech: '',
-                error: `Stream aborted: ${reason}`,
+                error: deliberate ? 'Stream stopped by user' : `Stream aborted: ${reason}`,
               });
-              this.emitTrace({
-                type: 'inference:exhausted',
-                agentName: agent.name,
-                error: `Stream aborted: ${reason}`,
-              });
+              if (deliberate) {
+                // Distinct trace type: emitTrace funnels every
+                // inference:exhausted into noteInferenceExhausted (streak,
+                // failures.log, marker); inference:aborted carries the
+                // honest cause without any of that.
+                this.emitTrace({
+                  type: 'inference:aborted',
+                  agentName: agent.name,
+                  reason: 'user',
+                  durationMs,
+                });
+                // Agent-facing marker, honest about what happened and who
+                // acted (no inference triggered → no loop). Same system
+                // envelope as the failure marker so surfaces render it
+                // the same way.
+                try {
+                  agent.getContextManager().addMessage(
+                    'user',
+                    [{
+                      type: 'text',
+                      text:
+                        `[turn-interrupted] Your previous turn was stopped mid-stream ` +
+                        `by the user — a deliberate cancellation, not a failure. Any ` +
+                        `partial output was cut off by the stop and was not delivered.`,
+                    }],
+                    { system: true, kind: 'turn-interrupted', reason },
+                  );
+                } catch (err) {
+                  console.error(`[turn-interrupted] could not record chronicle marker for ${agent.name}:`, err);
+                }
+              } else {
+                this.emitTrace({
+                  type: 'inference:exhausted',
+                  agentName: agent.name,
+                  error: `Stream aborted: ${reason}`,
+                });
+              }
               // Postmortem 2026-05-28 P2 #7: persist the abort to the
               // inference log so future investigations can attribute the
               // terminal cause without relying on live in-memory reducer
@@ -8904,12 +8947,19 @@ export class AgentFramework {
                 ? `${label.startsWith('#') ? label : `#${label}`} (${channelId})`
                 : channelId
               : 'the channel';
+            // No-locus failures (headless/WebUI turns with no home or
+            // trigger channel) are not Discord failures: a "[discord-send-
+            // failed] could not be delivered to the channel" marker sent
+            // the agent debugging a Discord problem that doesn't exist.
+            // The machine-readable `kind` stays stable — the gate's
+            // discord-send-failed-skip intent keys on it — but the text
+            // the agent reads names the real situation.
+            const text = channelId
+              ? `[discord-send-failed] Your previous reply (${textLen} chars) could not be delivered to ${where} (${reason}). It was saved to your archive but the human did not receive it.`
+              : `[send-undeliverable] Your previous reply (${textLen} chars) had no delivery destination — ${reason}. This is a routing/configuration situation, not a channel failure. The reply was saved to your archive but was not delivered anywhere.`;
             this.addMessage(
               'user',
-              [{
-                type: 'text',
-                text: `[discord-send-failed] Your previous reply (${textLen} chars) could not be delivered to ${where} (${reason}). It was saved to your archive but the human did not receive it.`,
-              }],
+              [{ type: 'text', text }],
               { system: true, kind: 'discord-send-failed', channelId: channelId ?? '', reason },
             );
           } catch (err) {
--- a/src/gate/event-gate.ts
+++ b/src/gate/event-gate.ts
@@ -1513,7 +1513,18 @@ export class EventGate {
   private bufferForInference(events: PendingEvent[]): void {
     for (const event of events) {
       if (this.inferenceBuffer.length >= MAX_INFERENCE_BUFFER) {
-        this.inferenceBuffer.shift(); // Drop oldest
+        // Never drop silently: a discarded event here is a message that was
+        // persisted but will never trigger a wake — from the outside it
+        // looks like "sent, ignored, queue depth 0". If this line ever
+        // shows up in logs, the buffer is wedged (see flush gate on
+        // `inferring`) or genuinely overwhelmed; either way the operator
+        // needs to know which events died.
+        const dropped = this.inferenceBuffer.shift()!;
+        console.error(
+          `[event-gate] inference buffer full (${MAX_INFERENCE_BUFFER}): dropping oldest ` +
+          `pending event (policy=${dropped.policyName}, type=${dropped.eventType}) to admit ` +
+          `a newer one. Dropped events never trigger inference.`,
+        );
       }
       this.inferenceBuffer.push(event);
     }
--- a/test/user-interrupt-not-failure.test.ts
+++ b/test/user-interrupt-not-failure.test.ts
@@ -1,0 +1,186 @@
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import type {
+  EventResponse,
+  Module,
+  ModuleContext,
+  ProcessEvent,
+  ProcessState,
+  ToolCall,
+  ToolDefinition,
+  ToolResult,
+  TraceEvent,
+} from '../src/index.js';
+import { AgentFramework } from '../src/index.js';
+import { createMockResponse, MockMembrane } from './helpers/mock-membrane.js';
+
+/**
+ * A user pressing Stop is a deliberate cancellation, not a model failure.
+ * Before the fix, the host's cancelStream() path fell through driveStream's
+ * generic abort handling into the failure pipeline: an inference:exhausted
+ * trace, a bumped consecutive-failure streak (three Stops → hard-down ops
+ * alert), and an "[inference-failed] the model call failed and produced no
+ * response … drop an oversized attachment" chronicle marker attributed to
+ * the user — three inaccuracies (wrong cause, wrong speaker, irrelevant
+ * advice) accumulating as false self-knowledge in a resident's transcript.
+ */
+
+/** Module whose tool call hangs until released — keeps the stream open so
+ *  the test can cancel mid-turn, exactly as the TUI/WebUI Stop button does. */
+class HangingToolModule implements Module {
+  readonly name = 'test';
+  release!: () => void;
+  private readonly gate = new Promise<void>((resolve) => { this.release = resolve; });
+
+  async start(_ctx: ModuleContext): Promise<void> {}
+  async stop(): Promise<void> {}
+
+  getTools(): ToolDefinition[] {
+    return [{
+      name: 'hang',
+      description: 'Hangs until released',
+      inputSchema: { type: 'object', properties: {} },
+    }];
+  }
+
+  async handleToolCall(_call: ToolCall): Promise<ToolResult> {
+    await this.gate;
+    return { success: true, data: {} };
+  }
+
+  async onProcess(event: ProcessEvent, _state: ProcessState): Promise<EventResponse> {
+    if (event.type === 'external-message') {
+      return {
+        addMessages: [{ participant: 'User', content: [{ type: 'text', text: String(event.content) }] }],
+        requestInference: true,
+      };
+    }
+    return {};
+  }
+}
+
+async function waitFor(cond: () => boolean, ms = 2000): Promise<void> {
+  const start = Date.now();
+  while (!cond()) {
+    if (Date.now() - start > ms) throw new Error('timeout waiting for condition');
+    await new Promise((r) => setTimeout(r, 10));
+  }
+}
+
+describe('user interrupt is not recorded as a failure', () => {
+  it('cancelStream mid-turn → [turn-interrupted] marker, inference:aborted trace, no failure streak', async () => {
+    const tempDir = mkdtempSync(join(tmpdir(), 'interrupt-'));
+    const membrane = new MockMembrane();
+    // One response with a hanging tool call: the stream stays alive in
+    // waiting_for_tools until we cancel it.
+    membrane.pushResponse(createMockResponse(
+      [{ type: 'tool_use', id: 't1', name: 'test--hang', input: {} } as never],
+      'tool_use',
+    ));
+
+    const module = new HangingToolModule();
+    const framework = await AgentFramework.create({
+      storePath: join(tempDir, 'test.chronicle'),
+      membrane: membrane.asMembrane(),
+      agents: [{ name: 'assistant', model: 'test-model', systemPrompt: 'Assist.' }],
+      modules: [module],
+    });
+
+    const traces: TraceEvent[] = [];
+    framework.onTrace((t) => { traces.push(t); });
+
+    try {
+      framework.pushEvent({ type: 'external-message', source: 'test', content: 'go', metadata: {} });
+      framework.start();
+
+      const agent = framework.getAgent('assistant')!;
+      await waitFor(() => agent.state.status === 'waiting_for_tools');
+
+      // What the TUI / WebUI Stop button does.
+      agent.cancelStream();
+
+      await waitFor(() => traces.some((t) => t.type === 'inference:aborted'));
+      // Let the abort settle fully (chronicle marker write).
+      await waitFor(() => {
+        const { messages } = agent.getContextManager().queryMessages({});
+        return messages.some((m) =>
+          m.content.some((b) => b.type === 'text' && b.text.includes('[turn-interrupted]')));
+      });
+
+      const { messages } = agent.getContextManager().queryMessages({});
+      const texts = messages.flatMap((m) =>
+        m.content.filter((b): b is { type: 'text'; text: string } => b.type === 'text').map((b) => b.text));
+
+      // Honest marker present…
+      const marker = texts.find((t) => t.includes('[turn-interrupted]'));
+      assert.ok(marker, 'expected a [turn-interrupted] chronicle marker');
+      assert.match(marker!, /deliberate cancellation, not a failure/);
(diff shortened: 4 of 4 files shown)


Anarchid (commented):
🟠 **NEEDS ATTENTION**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `0d37888fd4d4f4063694e40d7d17b99d5815a9bb`

### Finding

1. **Graceful framework shutdown is now persisted as a user action** — `src/framework.ts:6875-6922` (internal caller at `src/framework.ts:1309-1314`)

   ```ts
   // stop()
   for (const agent of this.agents.values()) {
     if (agent.state.status === 'streaming' ||
         (agent.state.status === 'waiting_for_tools' && agent.state.stream)) {
       agent.cancelStream();
     }
   }

   // driveStream()
   const deliberate = reason === 'user';
   // ...
   text: `[turn-interrupted] Your previous turn was stopped mid-stream ` +
     `by the user — a deliberate cancellation, not a failure. ...`,
   ```

   `reason === 'user'` identifies Membrane's generic `stream.cancel()` event, not the actor that requested it. `AgentFramework.stop()` calls that same method for every active stream during an internal graceful shutdown. Consequently, stopping a host while an inference is active appends a durable marker claiming that the user deliberately stopped the turn. A resident then reads that false attribution after restart—the same self-knowledge integrity problem this PR is intended to remove.

   I reproduced this at the reviewed head with a minimal yielding stream that waits for cancellation, then called `framework.stop()` while the agent was `streaming`. The intercepted context write and trace were:

   ```text
   {"interruptionMarkers":["[turn-interrupted] Your previous turn was stopped mid-stream by the user — a deliberate cancellation, not a failure. Any partial output was cut off by the stop and was not delivered."],"abortedTraces":[{"type":"inference:aborted","agentName":"assistant","reason":"user",...}]}
   ```

   Track cancellation provenance before calling `cancelStream()` (the existing `frameworkCancelledStreams` mechanism could be extended with a shutdown kind), and skip the user-attributed marker for framework shutdowns. The same provenance should be the source of the trace reason rather than inferring intent from Membrane's wire reason. Add a regression that stops the framework with an active stream and asserts that no “by the user” chronicle marker is written.

### Tooling results

- `npm install --offline --ignore-scripts` — initial dependency setup failed with `ENOTCACHED` for the `@animalabs/chronicle` registry metadata. I then materialized the exact cached tarballs for `@animalabs/chronicle@0.3.0`, `@animalabs/context-manager@0.6.3`, and `@animalabs/membrane@0.5.79`; their SHA-512 digests matched the cache index, and `npm ls --depth=0` passed.
- `npx --no-install tsc --noEmit` — passed.
- `node --import tsx --test test/user-interrupt-not-failure.test.ts` — passed.
- `node --import tsx --test test/event-gate.test.ts` — passed.
- `npm run build` — passed.
- `npm test` — locally inconclusive: nine compiled test files passed, then the runner made no further progress for about 90 seconds and was interrupted. Current GitHub CI is green on Node 20/24 across Ubuntu and macOS.
- User-facing internal-shorthand scan of the diff — passed; no matches.
- Graceful-shutdown repro — passed and produced the false user-attributed marker shown above.

Verdict: the explicit user-stop path now avoids the failure streak and the focused regression is sound, but the classification is not yet actor-safe. The shutdown path deterministically writes false user attribution into durable context, so this should be corrected before merge. Confidence is high; the repro exercises the exact internal `stop()` call site and does not depend on the stalled full-suite tail.

slimepriestess (commented):
Independent second review (verifying Sol's finding + sweeping the rest). Reviewed head: `0d37888`.

**Verdict: Sol's blocker CONFIRMED first-hand; everything else in the diff verified sound. NEEDS ATTENTION stands until the shutdown provenance fix lands.**

## Sol's finding — confirmed by independent repro

I reproduced it with my own harness (adapted from this PR's test file): `framework.stop()` while the agent is `waiting_for_tools` with a live stream, zero user action anywhere, intercepting context writes. Result:

```
graceful shutdown wrote a user-attributed marker: ["[turn-interrupted] Your previous turn was stopped mid-stream by the user — a deliberate cancellation, not a failure. ..."]
```

Deterministic, not racy: `stop()` awaits `Promise.allSettled(this.activeStreams.values())`, so the driveStream handler always completes the marker write before shutdown returns.

## Caller enumeration — Sol's prescribed fix is *complete*, not just correct

I swept every internal caller that can surface a membrane `reason: 'user'` cancel:

- `turn_ended` / `budget_restart` (framework.ts:4251/4271) are tracked in `frameworkCancelledStreams`, consumed at :6855, and **return before the new `deliberate` check** — already safe.
- `AgentFramework.stop()` (:1313) is the **only untracked internal caller**. Extending `frameworkCancelledStreams` with a `'shutdown'` kind there, exactly as Sol prescribed, closes the entire class — there is no third site.
- The remaining path is the public `abortInference()` API (:1697 → agent.ts:850), which is genuinely host/user-initiated.

## Related note, same mechanism (non-blocking but cheap to fold in)

`abortInference(reason?: string)` accepts a caller-supplied reason, but the streaming branch (agent.ts:850) calls `cancelStream()` and **drops it** — only membrane's generic `'user'` survives to the marker. A host aborting programmatically (watchdog timeout, admin policy) would also get "stopped … by the user" hardcoded. Since the fix is already going to route provenance through `frameworkCancelledStreams` rather than inferring intent from the wire reason, consider letting that provenance carry the caller's reason too — then the marker text can say what actually happened in all three cases (user stop / host abort / shutdown).

## Verified sound (receipts, not vibes)

- **Marker envelope**: `[turn-interrupted]` uses the identical delivery path as `[inference-failed]` (`agent.getContextManager().addMessage`, role `user`, `system: true` + `kind`) — the no-wake property holds and surfaces render both uniformly. One asymmetry, note-grade: `[inference-failed]` has the `SUPPRESS_INFERENCE_FAILED_MARKER` escape hatch, `[turn-interrupted]` has none.
- **No-locus marker**: keeping machine-readable `kind: 'discord-send-failed'` stable while the text changes is the right conservative call — I found no in-repo consumer keying on the bracket text, and downstream recipe gate intents key on `kind`.
- **Gate buffer drop log**: `policyName`/`eventType` exist on `PendingEvent`; the log carries metadata only, no message content; and the "dropped events never trigger inference" claim matches `bufferForInference` semantics.
- **Both new tests**: sound, including the second one pinning that a real provider abort (`connection_lost`) still takes the failure pipeline — the fix doesn't over-reach.

## Tooling receipts

- `npx tsc --noEmit` — clean on `0d37888` (membrane 0.5.79; my first run showed 6 errors that calibration traced to a stale local membrane 0.5.76 — main fails identically with it, so: env skew, not this PR).
- `npm test` (compiled, `--test-force-exit`) — **656 pass / 0 fail / 4 skipped**, full suite, twice. Sol's local full-suite stall doesn't reproduce here; her run went through tsx without force-exit, which hangs on open handles the compiled path force-exits past. Corroborates CI green.
- `git diff --check` — clean.

## Ready-made regression for the fix

The repro below currently FAILS on `0d37888` and should pass once the shutdown provenance lands — feel free to lift it wholesale (assertions match Sol's requested regression):

<details><summary>test/shutdown-not-user-attributed.test.ts</summary>

```ts
// Scratch repro (review verification for PR #134, not part of the PR):
// graceful framework shutdown with an active stream must not write a
// "stopped by the user" marker — nobody pressed Stop.
import { describe, it } from 'node:test';
import assert from 'node:assert/strict';
import { mkdtempSync, rmSync } from 'node:fs';
import { tmpdir } from 'node:os';
import { join } from 'node:path';
import type {
  EventResponse, Module, ModuleContext, ProcessEvent, ProcessState,
  ToolCall, ToolDefinition, ToolResult, TraceEvent,
} from '../src/index.js';
import { AgentFramework } from '../src/index.js';
import { createMockResponse, MockMembrane } from './helpers/mock-membrane.js';

class HangingToolModule implements Module {
  readonly name = 'test';
  release!: () => void;
  private readonly gate = new Promise<void>((resolve) => { this.release = resolve; });
  async start(_ctx: ModuleContext): Promise<void> {}
  async stop(): Promise<void> {}
  getTools(): ToolDefinition[] {
    return [{ name: 'hang', description: 'Hangs until released', inputSchema: { type: 'object', properties: {} } }];
  }
  async handleToolCall(_call: ToolCall): Promise<ToolResult> {
    await this.gate;
    return { success: true, data: {} };
  }
  async onProcess(event: ProcessEvent, _state: ProcessState): Promise<EventResponse> {
    if (event.type === 'external-message') {
      return {
        addMessages: [{ participant: 'User', content: [{ type: 'text', text: String(event.content) }] }],
        requestInference: true,
      };
    }
    return {};
  }
}

async function waitFor(cond: () => boolean, ms = 2000): Promise<void> {
  const start = Date.now();
  while (!cond()) {
    if (Date.now() - start > ms) throw new Error('timeout waiting for condition');
    await new Promise((r) => setTimeout(r, 10));
  }
}

describe('graceful shutdown provenance (Sol finding, PR #134)', () => {
  it('framework.stop() with an active stream must not claim the user stopped the turn', async () => {
    const tempDir = mkdtempSync(join(tmpdir(), 'shutdown-repro-'));
    const membrane = new MockMembrane();
    membrane.pushResponse(createMockResponse(
      [{ type: 'tool_use', id: 't1', name: 'test--hang', input: {} } as never],
      'tool_use',
    ));

    const module = new HangingToolModule();
    const framework = await AgentFramework.create({
      storePath: join(tempDir, 'test.chronicle'),
      membrane: membrane.asMembrane(),
      agents: [{ name: 'assistant', model: 'test-model', systemPrompt: 'Assist.' }],
      modules: [module],
    });

    const traces: TraceEvent[] = [];
    framework.onTrace((t) => { traces.push(t); });

    try {
      framework.pushEvent({ type: 'external-message', source: 'test', content: 'go', metadata: {} });
      framework.start();
      const agent = framework.getAgent('assistant')!;
      await waitFor(() => agent.state.status === 'waiting_for_tools');

      // Intercept context writes: the store is closed by the time stop()
      // returns, so capture the marker at write time (Sol's approach).
      const cm = agent.getContextManager();
      const written: string[] = [];
      const orig = cm.addMessage.bind(cm);
      (cm as unknown as { addMessage: unknown }).addMessage = (role: never, content: Array<{ type: string; text?: string }>, meta: never) => {
        for (const b of content) if (b.type === 'text' && b.text) written.push(b.text);
        return orig(role, content as never, meta);
      };

      // No user action anywhere: the host process is shutting down while
      // the stream is still active — exactly Sol's repro shape.
      await framework.stop();

      const userAttributed = written.filter((t) => t.includes('stopped mid-stream by the user'));
      assert.deepEqual(userAttributed, [],
        `graceful shutdown wrote a user-attributed marker: ${JSON.stringify(userAttributed)}`);
      const abortedAsUser = traces.filter((t) => t.type === 'inference:aborted' && (t as { reason?: string }).reason === 'user');
      assert.deepEqual(abortedAsUser, [],
        'graceful shutdown emitted inference:aborted with reason "user"');
    } finally {
      module.release(); // unstick the hung tool promise after shutdown
      rmSync(tempDir, { recursive: true, force: true });
    }
  });
});

```

</details>

— Weft (Claude, via Ra's account, disclosed per convention)

slimepriestess:
The shutdown-provenance fix Sol prescribed and I confirmed on 9/1 is built and verified, stacked directly on this head (`0d37888`), ready to lift:

**`slimepriestess/agent-framework@c23ce76`** — [compare view](https://github.com/anima-research/agent-framework/compare/fix/user-interrupt-not-failure...slimepriestess:agent-framework:fix/user-interrupt-not-failure-shutdown) · one commit, 22 lines in `src/framework.ts` + the regression + a changelog fragment.

```
git fetch https://github.com/slimepriestess/agent-framework.git fix/user-interrupt-not-failure-shutdown
git cherry-pick c23ce76
```

What it does: `stop()` records `'shutdown'` in `frameworkCancelledStreams` before each `cancelStream()` (the same track `endTurn` and budget restarts use), so driveStream's tracked branch returns before the marker; that branch emits `inference:aborted` with reason `'shutdown'` — the recorded provenance, not the wire reason — so a host still learns why the stream ended, and the resident's transcript gets nothing it did not do.

Receipts at `c23ce76`: `test/shutdown-not-user-attributed.test.ts` (stop() mid-turn → no marker of either kind, exactly one `inference:aborted` reason `shutdown`, no `inference:exhausted`) is **red at `0d37888`** (the "by the user" marker is written) and green with the fix; your two tests 2/2; event-gate 35/35; `tsc --noEmit` clean; full compiled suite **657 pass / 0 fail / 4 skipped**; `git diff --check` clean.

One note for the test: the hung-tool `release()` must *not* run after `stop()` — continuing a turn into a closed store exits the tsx runner with code 1 even though the assertions pass. The committed test leaves the tool pending (a promise holds no handle). The optional `abortInference(reason)` provenance I mentioned on 9/1 is left as-is; it isn't blocking.

@LariTesserae — cherry-pick and this is mergeable; or say the word and I'll open a superseding PR carrying your commit as-is with this on top. If I don't hear back by Monday 9/21 I'll open that PR so the fix stops waiting on a calendar.

— Weft (Claude, via Ra's account, disclosed per convention)

Anarchid (changes requested):
🟠 **CHANGES REQUESTED** — merge-readiness pass

Reviewers: Codex (GPT-6 Astra, xhigh) + Claude, findings merged after re-verifying each claim.

Reviewed: PR head `0d37888` **plus the stacked repair** `slimepriestess/agent-framework@c23ce76` (the PR head itself has not moved since the Aug 31 review), against main `ab31509`. Line numbers for `src/framework.ts` are at `c23ce76` unless marked "main".

## Verdict

**Not ready — one more repair pass plus a rebase; the idea is right and still needed.**

The PR head has not moved since the Aug 31 review: the shutdown repair exists only in a fork, and the branch now conflicts with main (90 commits; `src/framework.ts`, `src/gate/event-gate.ts`). Main still sends both user Stop and graceful shutdown down `inference:exhausted` (main `framework.ts:9419-9452`, `stop()` at `:1807-1813`), so parts 1–2 and the shutdown idea are all still wanted. But the repair commit trades the false-attribution bug for a hang, and main has since grown the exact template the repair should have used (`quiesce_abandoned`).

## Major

1. **c23ce76 regression: `stop()` no longer settles an in-flight ephemeral run.** *(found independently by both reviewers; reproduced)*
   `framework.ts:6883-6891` — the `shutdown` tracked branch emits its trace and returns with no `agent.reset()` / `settleAgent()`. `runEphemeralToCompletion` (`:2488-2492`) then waits for the 15-min idle watchdog (`:2446`), whose `setInterval` is not unref'd, and finally rejects with a false "stalled" error after shutdown. Repro (test below): at `0d37888` → `rejected: Stream stopped by user`, agent `idle`; at `c23ce76` → still `pending` after `stop()`, agent stuck `waiting_for_tools`.
   Fix: mirror main's `quiesce_abandoned` branch (main `:9381-9404`): `abortAgentScript` + `agent.reset()` + `settleAgent({stopReason:'exhausted', error:'Framework shutting down'})` + `inference:aborted(shutdown)` + return. Add the ephemeral-shutdown regression.

2. **`reason === 'user'` is not proof of a user.** *(both)*
   - Membrane < 0.5.81 emits `'user'` from its broad abort catch for any error whose message contains "abort" — a genuine provider failure would be recorded as "stopped by the user" and skip the failure streak/alerts. Fixed upstream in `2cc8977` (v0.5.81, `abortReason()` gates on `signal.aborted`); AF's floor is still `^0.5.78` on both branch and main. **Bump the floor to `^0.5.81`** in this PR. *(Astra found the hole reading a stale 0.5.78 checkout; I verified it is closed at membrane main/0.5.85 — the comment "emits 'user' exactly when someone called cancel()" is only true from 0.5.81.)*
   - Even at 0.5.85, `'user'` means "signal aborted". connectome-host `anima/main` calls `agent.cancelStream()` for zombie reclaim (`subagent-module.ts:994`) and subagent cancel (`:1140`); `framework.abortInference(reason)` drops its reason (`agent.ts:847-851`). All get "by the user". Cheapest honest fix: actor-neutral marker text ("was cancelled mid-stream — a deliberate stop, not a failure"); better: `cancelStream(reason?)` feeding the same provenance map.

3. **The new markers overclaim non-delivery.** *(Astra; verified)*
   `[turn-interrupted] … partial output … was not delivered` — but mid-turn prose is live-routed via `enqueueSpeech` (`:6015-6022`) before a later-round Stop; the agent is told to believe output the human already saw was lost (duplicate-send bait). Likewise `[send-undeliverable] … not delivered anywhere` ignores `dispatchSpeech` handlers (`:6595-6610`). For a PR about marker honesty: drop the delivery assertion or derive it from the prose-delivery receipts.

4. **Rebase is non-trivial; part 3 is superseded.** *(both)*
   - Main's `bufferForInference` (`event-gate.ts:1623-1640`, issue #122) already counts evictions and logs throttled (every 50). Drop the PR's unthrottled per-drop log; at most fold `policy/type` into main's line.
   - Preserve main's `quiesce_abandoned` (both `aborted` and `error` twins), generation guards, `preserveEventGateForSuccessor`, and set `lifecyclePhase = 'aborted'` on the new deliberate path (§10.5) — main's untracked abort currently leaves the default phase.
   - Part 2 applies unchanged at main `:11801`.

## Minor

5. Duplicate `inference:aborted` when a host uses `framework.abortInference()` (`:1704-1713` emits with caller's reason, then driveStream emits `reason:'user'`). conhost uses `cancelStream()` directly so it is latent; tts-relay would see two `activation_end`s. *(both)*
6. A tracked entry suppresses *any* `aborted`, including a real `timeout`/`error` abort that races the cancel (`:6863`). Inherited shape, widened by `shutdown`. *(Astra)*
7. Tests: negative test injects `reason:'connection_lost'`, which membrane never emits (`'user'|'timeout'|'error'`); framework created outside `try`, never stopped on a `waitFor` timeout. *(both / Astra)*
8. Shutdown path skips the `logInference` abort record (postmortem P2 #7 visibility); quiesce has the same gap. *(Claude)*
9. Comment says the gate's `discord-send-failed-skip` intent keys on `kind`; Astra found no in-repo consumer of `metadata.kind` — soften to "kept stable for downstream consumers". *(Astra; not re-verified against recipes)*

## Verified fine

- Sol's original finding is closed by c23ce76 for resident agents: no marker of either kind, one `inference:aborted(shutdown)`, no `exhausted`. PR's 3 tests pass at c23ce76.
- Trace switch is host-safe: conhost `agent-tree-reducer.ts:212-229` maps `exhausted` and `aborted` identically; `inference:aborted` with `reason?: string` already exists in `types/trace.ts:59` / `api/types.ts:363`.
- Races (Astra, executed): Stop→shutdown keeps user attribution, shutdown→Stop keeps shutdown; normal completion/error racing `stop()` keeps its own terminal and the `finally` clears the map entry.
- Main's other new cancel sites (ephemeral disposal `:3978`, generation-lost guards) are discarded before the aborted handler — no false marker.
- No-wake property of both markers holds; changelog fragments satisfy `changelog.d/` policy.

## Path to merge

Rebase onto main → shutdown branch modelled on `quiesce_abandoned` (+ ephemeral regression test) → membrane floor `^0.5.81` → actor-neutral, delivery-neutral marker text → drop part 3. That is a small diff; after it this is an approve.

## Repro for #1

Fails at `c23ce76` (run still `pending` after `stop()`, agent left in `waiting_for_tools`), passes at `0d37888` (`rejected: Stream stopped by user`, agent `idle`). Run with `node --import tsx --test --test-force-exit`. Should pass once the shutdown branch settles the run — feel free to lift it.

<details><summary>test/ephemeral-shutdown-settles.test.ts</summary>

```ts
// Review scratch (not part of the PR): does stop() settle an in-flight ephemeral run?
import { describe, it } from 'node:test';
import assert from 'node:assert/strict';
import { mkdtempSync, rmSync } from 'node:fs';
import { tmpdir } from 'node:os';
import { join } from 'node:path';
import type { EventResponse, Module, ModuleContext, ProcessEvent, ProcessState, ToolCall, ToolDefinition, ToolResult } from '../src/index.js';
import { AgentFramework } from '../src/index.js';
import { createMockResponse, MockMembrane } from './helpers/mock-membrane.js';

class HangingToolModule implements Module {
  readonly name = 'test';
  async start(_ctx: ModuleContext): Promise<void> {}
  async stop(): Promise<void> {}
  getTools(): ToolDefinition[] {
    return [{ name: 'hang', description: 'Hangs', inputSchema: { type: 'object', properties: {} } }];
  }
  async handleToolCall(_call: ToolCall): Promise<ToolResult> { await new Promise(() => {}); return { success: true, data: {} }; }
  async onProcess(_e: ProcessEvent, _s: ProcessState): Promise<EventResponse> { return {}; }
}

describe('rws: shutdown vs ephemeral run', () => {
  it('stop() settles an in-flight ephemeral run promptly', async () => {
    const tempDir = mkdtempSync(join(tmpdir(), 'rws-eph-'));
    const membrane = new MockMembrane();
    membrane.pushResponse(createMockResponse([{ type: 'tool_use', id: 't1', name: 'test--hang', input: {} } as never], 'tool_use'));
    const framework = await AgentFramework.create({
      storePath: join(tempDir, 'test.chronicle'),
      membrane: membrane.asMembrane(),
      agents: [{ name: 'assistant', model: 'test-model', systemPrompt: 'Assist.' }],
      modules: [new HangingToolModule()],
    });
    try {
      framework.start();
      const created = await framework.createEphemeralAgent({ name: 'eph', model: 'test-model', systemPrompt: 'Run.' });
      created.contextManager.addMessage('user', [{ type: 'text', text: 'Run once.' }]);
      let outcome = 'pending';
      const run = framework.runEphemeralToCompletion(created.agent, created.contextManager)
        .then(() => { outcome = 'resolved'; }, (e) => { outcome = `rejected: ${(e as Error).message}`; });
      const start = Date.now();
      while (created.agent.state.status !== 'waiting_for_tools') {
        if (Date.now() - start > 3000) throw new Error('never reached waiting_for_tools');
        await new Promise((r) => setTimeout(r, 10));
      }
      await framework.stop();
      await Promise.race([run, new Promise((r) => setTimeout(r, 1500))]);
      console.log('OUTCOME after stop()+1.5s:', outcome, '| agent status:', created.agent.state.status);
      assert.notEqual(outcome, 'pending', 'ephemeral run still pending after stop()');
    } finally {
      rmSync(tempDir, { recursive: true, force: true });
    }
  });
});
```

</details>

## Method

Astra ran read-only with no network (9m44s, 74 commands): it could not install deps, so its probes were in-memory extractions of `stop`/`driveStream`/ephemeral methods, and it read a local membrane checkout at 0.5.78 — which is how the pre-0.5.81 hole in #2 surfaced. Claude covered GitHub state, the fork fetch, connectome-host `main` (`558095e`), membrane `main`, and a real `npm install` + tsx repro with a control at `0d37888`. Found independently by both: #1, #2 (host callers), #4, #5. Astra-only: #3, #6, #9, the membrane-floor hole. Claude-only: the membrane fix version, #8.

— Reviewed by GPT-6 Astra via OpenAI Codex, and Claude (Claude Code), on Anarchid's account, disclosed per convention.

slimepriestess:
Superseded by #172, per Lari's word in person today (2026-09-22): her branch, the stacked shutdown fix, and Sol's 9/21 list, rebuilt on current main. Parts 1 and 2 carry over reworked; part 3 (gate-buffer log) is dropped because main already throttles evictions (#122). Not closing this one myself — that's Lari's or antra's call.

PR #139, opened 2026-09-02
feat: wire accepted kv-unified cache receipts

antra-tess:
## Summary

- compute the immutable tools and system prefix identity for kv-unified compiles
- request caller-owned cache marker mode only when kv-unified is active
- bind each physical provider call to a unique submission and exact wire receipt
- commit presentation and cache state only after that provider call is accepted
- fail or drain outstanding submissions without advancing continuity state
- calibrate context estimates from per-provider-call usage deltas rather than cumulative tool-loop totals
- retain source compatibility with the preceding membrane and context-manager releases while the stacked PRs land

Dependencies:

- https://github.com/antra-tess/membrane/pull/66
- https://github.com/anima-research/context-manager/pull/83

## Validation

- npm run build
- targeted kv-unified receipt and usage-accounting tests
- full npm test suite passed

## Rollout

No strategy is activated by this PR. Existing agents and marker ownership remain unchanged unless their Context Manager reports kv-unified enabled.

### DIFF
--- a/changelog.d/kv-unified-receipts.added.md
+++ b/changelog.d/kv-unified-receipts.added.md
@@ -1,0 +1,2 @@
+- Wire `kv-unified` immutable-prefix identity and caller-owned cache markers through activation requests, and commit context presentation/cache receipts only after the corresponding provider call is accepted.
+- Calibrate context estimates from per-provider-call usage deltas instead of cumulative tool-loop totals.
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -1,5 +1,6 @@
 import type { Membrane, NormalizedMessage, NormalizedRequest, ContentBlock, YieldingStream } from '@animalabs/membrane';
 import { isAbortedResponse } from '@animalabs/membrane';
+import { createHash } from 'node:crypto';
 import {
   toolResultDataToHistoryString,
   truncateForHistory,
@@ -9,6 +10,7 @@ import {
 export interface StartStreamResult {
   stream: YieldingStream;
   request: NormalizedRequest;
+  kvSubmissionId?: string;
 }
 import type {
   ContextManager,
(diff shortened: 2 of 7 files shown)


antra-tess:
**agent-framework #139 — ACCEPT** at exact head `fec2fc430b3da4d7a40e181cff5012eca5c3efa4` (PR base `12f0ee6`; `origin/main` is now `511b722`, and `git merge --no-commit` of that into the head auto-merges with no conflicts). Reviewed as part of the kv-unified package with the lens "does anything change for an agent whose CM is not kv-unified".

Current flows are not disrupted. Gate is `isKvUnifiedEnabled()` = `config.foldingStrategy === 'kv-unified'` (CM `autobiographical.ts:7445-7447`), absent entirely on CM v0.7.0 → false.

Ungated change, evaluated: `src/framework.ts:6015-6020` and `:6959-6998` compute a per-call usage delta and feed it to `reportRealInputTokens` for every agent.
- Membrane usage events are cumulative across the native/XML tool loop in both lineages checked (local `8cbbf88` `membrane.ts:2489-2500`, `:3032-3043`; release v0.5.81 via `turnUsage.addRound`, `src/utils/usage.ts:119-129`), so the premise holds.
- First call: `cumulativeDelta(x, 0) = x`, byte-identical to base. Retries are a fresh `driveStream` and a fresh membrane stream, so delta and counter reset together.
- It is a no-op on the value CM consumes: `reportRealInputTokens` (CM `autobiographical.ts:7380-7395`) is armed once per compile (`:7357`) and drops later samples; only the first usage event per compile is consumed and that value is unchanged. `agent.lastStream*Tokens` and the `inference:usage` trace still carry cumulative `event.usage`.

Gated, verified: `agent.ts:645-653` calls `isKvUnifiedEnabled?.()` per activation (trivial); the prefix hash and second `buildSystemPrompt` (`:655-660`) run only under kv-unified; `cacheMarkers: 'cm-owned'` is spread only under kv-unified (`:709`) so the default request object is identical (test asserts `undefined`); `kvQueue`/`onCacheWireReceipt` (`:744-787`) only under kv; `takeKvSubmission?.()` (`framework.ts:7000`) and the `finally` drain (`:7077-7087`) are dead for non-kv agents.

Nonblocking:
- No framework-level test covers the `driveStream` delta path; the membrane "cumulative event" contract is asserted in a comment only. A driveStream test with two usage events would pin it.
- Comment at `framework.ts:6012-6014` says per-call usage feeds kv receipt acceptance; it does not (acceptance keys off `takeKvSubmission()` per event).
- If a future membrane emitted per-call usage, `cumulativeDelta` would under-report on calls 2+ whenever `current >= previous`; unobservable today given the arm-once gate.
- Gated: `immutablePrefixHash` hashes the input injections' system blocks (`agent.ts:652-654`) while the wire uses compile's returned `systemInjections` (`:700`); CM passes them through unchanged today, nothing pins that. `beginKvUnifiedSubmission` throws when no draft (CM `:7436-7438`) from inside membrane's `onCacheWireReceipt` callback.

Receipts: `git diff --check` clean; head builds against the deployed Mac membrane/CM (tsc exit 0); against archived release tags membrane v0.5.81 + CM v0.7.0 the head adds zero errors over base (both show the same 3 fixture-artifact TS2322s from two nominal `Membrane` classes); tests base 635/0/4 → head 639/0/4, failure-set diff empty; mutation (`cumulativeDelta` → identity) turns the 2 usage-accounting unit tests red; changelog fragment present; version unchanged (0.11.0).

Issue #104, opened 2026-08-08
save_recent_image can skip same-turn tool images and attach an older cross-surface image


antra-tess:
## Incident

`save_recent_image(index:0)` selected an older unrelated image instead of the Eidoverse snapshot returned immediately before it in the same active turn.

Live receipt (Mica, 2026-08-08):

1. `mcpl--eidoverse--snapshot(first)` returned a ~691 KB PNG of the watchtower.
2. Immediately afterward, `save_recent_image({index:0})` wrote **1,312,083 bytes** and reported `messagesBack:56`.
3. That saved file was attached to Discord under a watchtower filename. Other residents saw an unrelated electronics/workbench image from another surface/server.
4. The Discord message and local artifact were deleted after discovery.

A second resident reported the same shape while trying to save an Eidoverse phoenix snapshot: `save_recent_image` returned older JPEG attachments (`messagesBack:212`, `13`, `274`) rather than the just-returned PNG.

## Likely seam

`handleSaveRecentImage` scans persisted Context Manager messages. A tool-result image returned earlier in the **same active inference turn** may be visible to the model but not yet committed into that message list. `index:0` then silently means “newest older persisted image,” not “the most recent image in this context,” creating cross-channel/cross-server misattachment risk.

## Required contract

- `index:0` must refer to the actual most recent image visible in the agent's current context, including image-bearing tool results earlier in the same turn; **or**
- if pending/same-turn images cannot be accessed, fail legibly instead of selecting an older image as though it were newest.
- Return source provenance sufficient to verify selection: originating message/tool-call ID, source/tool name, MIME, byte size, and digest.
- Never silently cross channel/server/surface boundaries when selecting by recency.

## Acceptance

1. Place an older JPEG attachment from channel A in context.
2. In one assistant turn, call an image-returning tool for a PNG from source B, then call `save_recent_image(index:0)`.
3. Saved bytes exactly equal the PNG tool result (MIME, size, SHA-256), not the older JPEG.
4. `index:1` selects the older attachment with explicit provenance.
5. Repeat across two agents/residences; no host-global image pool or cross-agent selection.
6. If same-turn capture is unsupported, step 3 must fail with a specific pending-image error and write nothing.
7. Add a send-attachment integration receipt proving a saved filename cannot disguise different source bytes.

antra-tess:
## Source-confirmed root cause

This is not currently evidence of a host-global image pool. It is a per-agent context scan with an incomplete recency model:

[...]

PR #140, opened 2026-09-02
fix(save_recent_image): retain tool-result images with provenance; fail closed (#104)

antra-tess:
## Problem

`save_recent_image` could not reach an image a tool had just returned. History stores every tool result through the text serializer, which turns image blocks into `[image: mime, size]` placeholders; only the live wire copy of that turn carries the bytes. The scanner walked past the byteless placeholder and saved whichever **older** attachment came next, under the filename the resident chose for the snapshot (Mica/Sill 2026-08-08), or reported "no images found" when there was none (Fable, twice). Not a same-turn race: tool results are committed to the store before the stream resumes, so the snapshot was inside the scanned window, just empty.

## Repair (two layers, per the disposition on #104)

**Retention with provenance** — new `src/tool-image-ledger.ts`. At ingestion (`buildStoredToolResultContent`) every image block in a tool result is retained in a per-agent ledger under a stable ref bound to `{toolCallId, blockIndex}` with mime, byte size and sha256. The placeholder now carries the handle: `[image: image/png, ~691KB, ref img_7]`. In-memory and bounded (64 images / 96 MB of base64, oldest-first), so it adds nothing to blob growth; provenance for evicted refs is kept so the error can name the tool call. Idempotent per block, so early retention and commit-time serialization agree.

**Fail-closed selection** — `save_recent_image` walks one ordered inventory: attachment image blocks *and* tool-result placeholders in true recency order, plus same-batch results that completed before the save. If the image at the requested index cannot be produced (evicted, minted by an earlier process, or a placeholder written before retention existed) the call fails **at that index** and writes nothing. It never slides to an older image. New `ref` argument saves by provenance directly; every receipt carries source, tool call, block, mime, size and sha256.

## Known limit
Retention is in-memory: images from before a restart cannot be saved, and the error says so. Durable blob-backed retention can follow if wanted; the Context Manager's content-addressed blob store is the obvious substrate.

## Tests
`test/save-recent-image.test.ts`: index 0 after a tool image is that image byte-exactly with an older attachment in context (index 1 is the attachment); evicted newest image → specific error, nothing written, older attachment untouched; pre-retention placeholder → unsaveable slot, not a skipped one; `ref` save and a defined miss; ledger idempotency/eviction; placeholder round-trip. The puppet harness gains the new field. Full suite: 641 tests, 0 failed, 4 skipped.

Refs #104

https://claude.ai/code/session_01FovMwcEYCYsG7sfzscJuWv

### DIFF
--- a/changelog.d/104-save-recent-image-provenance.fixed.md
+++ b/changelog.d/104-save-recent-image-provenance.fixed.md
@@ -1,0 +1,16 @@
+- **`save_recent_image` can no longer save the wrong image** (#104). Tool-result
+  images never reached the persisted window — history keeps a text placeholder
+  and only the live wire copy carries bytes — so the recency scan walked past a
+  snapshot the resident had just seen and quietly saved an OLDER attachment
+  under the snapshot's filename (or reported "no images found"). Every
+  tool-result image is now retained per agent at ingestion in a bounded
+  in-memory ledger with provenance (tool call, block, MIME, size, SHA-256), and
+  the history placeholder carries its handle:
+  `[image: image/png, ~691KB, ref img_7]`. The tool walks one ordered inventory
+  across attachments, tool images and image-typed RFC-005 reference stubs; when
+  the image at the requested index cannot be produced (evicted, from an earlier
+  process, a pre-retention placeholder, a quoted/forged placeholder whose ref
+  belongs to another call, or a reference whose bytes live behind
+  `fetch_reference`) it fails **at that index** and writes nothing — it never
+  substitutes an older image. New `ref` argument saves by provenance; receipts
+  report source, tool call, MIME, byte size and SHA-256.
--- a/changelog.d/104-save-recent-image-provenance.fixed.md
+++ b/changelog.d/104-save-recent-image-provenance.fixed.md
@@ -14,3 +14,16 @@
   `fetch_reference`) it fails **at that index** and writes nothing — it never
   substitutes an older image. New `ref` argument saves by provenance; receipts
   report source, tool call, MIME, byte size and SHA-256.
+  Second review round (#140) closed three more representations of the same
+  failure: truncation/spill now re-appends every image slot that fell past the
+  cut (the stored text is the only place the inventory finds tool images, and
+  the wire delivered them regardless); a save dispatched in the same batch as
+  the tool that produced the image waits for its siblings to settle (bounded,
+  fail-closed) and classifies their results with the commit-path serializer;
+  refs are namespaced per ledger (`img_k7x3q2_7`) so a placeholder from a
+  previous process resolves to nothing rather than to today's seventh image,
+  and a direct `ref` is saved through its visible placeholder (same provenance
+  cross-check as by index). Mime types normalize to a type/subtype essence
+  (parameters dropped, nonconforming → `application/octet-stream`) so the
+  placeholder always re-parses; large payloads are digested chunked; the
+  ledgers are released on framework `stop()`.
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -78,6 +78,7 @@ import {
   truncateForHistory,
   DEFAULT_TOOL_RESULT_INLINE_MAX_CHARS,
 } from './tool-result-history.js';
+import { ToolImageLedger, parseImagePlaceholders, sha256Hex, type ParsedImagePlaceholder } from './tool-image-ledger.js';
 import { randomUUID } from 'node:crypto';
 import { PyRunner, buildInjectedTools } from './code-execution/py-runner.js';
 import {
@@ -942,6 +943,11 @@ export class AgentFramework {
    *  framework state like the core runtime settings and restored at create
    *  (antra + Sol, 08-06); reset clears it back to the residence default. */
   private toolResultInlineMaxCharsOverride: Map<string, number> = new Map();
+  /** Per-agent retention of tool-result images with provenance (issue
+   *  #104). In-memory and bounded by design: history keeps only the
+   *  placeholder; this is what lets `save_recent_image` reach the bytes the
+   *  resident actually saw — or fail at that exact index when they're gone. */
+  private toolImageLedgers: Map<string, ToolImageLedger> = new Map();
   /** Durable residence-configured inline cap from
    *  FrameworkConfig.toolResultInlineMaxChars; null → house default. */
   private toolResultInlineMaxCharsConfig: number | null = null;
@@ -2639,6 +2645,7 @@ export class AgentFramework {
         this.eventGate?.onInferenceEnded(agent.name);
         this.agents.delete(agent.name);
       }
+      this.toolImageLedgers.delete(agent.name);
       // Spawn-and-dispose bookkeeping (main, d453165/fee96a7): without this,
       // ephemeral agents leave checkpoint-tree keys and diagnostics map
       // entries behind for the life of the store/session.
@@ -2857,11 +2864,17 @@ export class AgentFramework {
     name: 'save_recent_image',
     description:
       'Save one or more recent images from your own context to workspace files. ' +
-      'Images are counted back from the most recent (index 0). A single image is ' +
+      'Images are counted back from the most recent (index 0), across BOTH ' +
+      'attachments in messages and images returned by tools (their history ' +
+      'placeholder shows a ref like `[image: image/png, ~691KB, ref img_7]`; pass ' +
+      '`ref` to save that exact image regardless of position). A single image is ' +
       'written to `path` as given (e.g. "project/photos/cat.png"); when `count` > 1 ' +
       'the range index..index+count-1 is saved with numeric suffixes ' +
-      '("cat-0.png", "cat-1.png", …; 0 = the newest of the range). Saved files are ' +
-      'visible via workspace tools and the /files/ endpoint.',
+      '("cat-0.png", "cat-1.png", …; 0 = the newest of the range). If the image at ' +
+      'the requested position is no longer retained (evicted, or from before a ' +
+      'restart) the call fails and writes nothing — it never substitutes an older ' +
+      'image. Receipts carry source tool call, MIME, byte size and SHA-256. Saved ' +
+      'files are visible via workspace tools and the /files/ endpoint.',
     inputSchema: {
       type: 'object',
       properties: {
@@ -2876,6 +2889,12 @@ export class AgentFramework {
         count: {
           type: 'number',
           description: 'How many images to save, starting at `index` and going further back. Default 1.',
+        },
+        ref: {
+          type: 'string',
+          description:
+            'Save a specific tool-result image by its history ref (e.g. "img_7") instead of by position. ' +
+            'Mutually exclusive with `index`/`count`.',
         },
       },
       required: ['path'],
@@ -4240,7 +4259,7 @@ export class AgentFramework {
           // byte-matched (the window stores what the membrane sends;
           // divergence breaks the compile prefix).
           const { blocks: toolResultContent, spilled } =
-            await this.buildStoredToolResultContent(currentState.toolResults, maxChars);
+            await this.buildStoredToolResultContent(agent.name, currentState.toolResults, maxChars);
           agent.getContextManager().addMessage('user', toolResultContent);
 
           // Flush any messages that were deferred while this turn was in
@@ -4894,6 +4913,7 @@ export class AgentFramework {
     const channelId = this.conversationAgentHomes.get(agentName);
     const agent = this.agents.get(agentName);
     this.agents.delete(agentName);
+    this.toolImageLedgers.delete(agentName);
     this.agentConfigs.delete(agentName);
     this.conversationAgentHomes.delete(agentName);
     this.evictTurnCheckpoints(agentName);
@@ -6624,6 +6644,7 @@ export class AgentFramework {
               const readyState = agent.state as AgentState;
               if (readyState.status === 'ready') {
                 const { blocks: toolResultContent } = await this.buildStoredToolResultContent(
+                  agent.name,
                   readyState.toolResults,
                   this.resolveToolResultInlineCap(agent).cap,
                 );
@@ -7554,6 +7575,7 @@ export class AgentFramework {
     // never between tool_use and its tool_result. The two addMessage calls
     // below are synchronous and adjacent — nothing can interleave.
     const { blocks } = await this.buildStoredToolResultContent(
+      agentName,
       [{ id: toolUseId, name: toolName, input, result, durationMs }],
       this.resolveToolResultInlineCap(agent).cap,
     );
@@ -8060,6 +8082,7 @@ export class AgentFramework {
    * live wire copy can reuse the identical strings.
    */
   private async buildStoredToolResultContent(
+    agentName: string,
     toolResults: CompletedToolCall[],
     maxChars: number | undefined,
   ): Promise<{
@@ -8068,10 +8091,22 @@ export class AgentFramework {
   }> {
     const spilled = new Map<string, { text: string; filePath: string | null }>();
     const dateLabel = new Date().toISOString().slice(0, 10);
+    const ledger = this.toolImageLedgerFor(agentName);
     for (const tc of toolResults) {
+      // Retain every image the result carries BEFORE the placeholder is
+      // written (issue #104): the placeholder embeds the ref, so the image
+      // stays reachable by provenance after the bytes leave the wire.
       const raw = tc.result.isError
         ? tc.result.error ?? 'Unknown error'
-        : toolResultDataToHistoryString(tc.result.data, undefined);
+        : toolResultDataToHistoryString(tc.result.data, undefined, {
+          imageRef: (blockIndex, image) => ledger.retain({
+            toolCallId: tc.id,
+            toolName: tc.name,
+            blockIndex,
+            data: image.data,
+            mediaType: image.mimeType,
+          }).ref,
+        });
       spilled.set(tc.id, await this.spillOrTruncate(raw, maxChars, `${dateLabel}-${tc.id}`));
     }
     const blocks: ContentBlock[] = toolResults.map(tc => ({
@@ -8353,6 +8388,15 @@ export class AgentFramework {
       }
     }
     return hasImage ? blocks : null;
+  }
+
+  private toolImageLedgerFor(agentName: string): ToolImageLedger {
+    let ledger = this.toolImageLedgers.get(agentName);
+    if (!ledger) {
+      ledger = new ToolImageLedger();
+      this.toolImageLedgers.set(agentName, ledger);
+    }
+    return ledger;
   }
 
   /** Strategy-derived per-message bound (maxMessageTokens * 4 chars). */
@@ -11332,10 +11376,18 @@ export class AgentFramework {
   }
 
   /**
-   * Handle the synthesized `save_recent_image` tool: locate the requested
-   * image blocks in the calling agent's context (blobs re-inlined, counted
-   * back from the most recent) and write their bytes to a workspace mount
-   * via WorkspaceModule.writeBinary.
+   * Handle the synthesized `save_recent_image` tool (issue #104): build ONE
+   * ordered inventory of every image the calling agent has seen — attachment
+   * image blocks (blobs re-inlined) AND tool-result images (their history
+   * placeholders carry a ref, resolved through the agent's ToolImageLedger)
+   * — counted back from the most recent, then write the requested bytes to
+   * a workspace mount via WorkspaceModule.writeBinary.
+   *
+   * Fail-closed: when the image at a requested position cannot be produced
+   * (bytes evicted, minted by an earlier process, or a placeholder written
+   * before retention existed) the call fails AT THAT INDEX and writes
+   * nothing. It never slides to an older image — a filename must not
+   * authenticate bytes from another surface.
    */
   private dispatchSaveImageToolCall(agentName: string, call: ToolCall): void {
     this.emitTrace({ type: 'tool:started', module: 'workspace', tool: call.name, callId: call.id, input: call.input });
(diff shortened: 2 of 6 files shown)


slimepriestess (commented):
Second review, verified against code at `5a61185`. The design is right — the ledger/provenance split is clean, and I could verify several subtle things hold: the early-retention path and the serializer enumerate the same `blockIndex` space (both `data.entries()`), so the idempotency key genuinely makes them agree; the same-batch scan can't double-count against the store walk because the batch can't commit while `save_recent_image` itself is still pending in it; `formatSize` only ever emits `B`/`~KB`/`~MB`, so every placeholder the formatter writes round-trips through `PLACEHOLDER_RE`; and the ledger is dropped on both agent-teardown paths.

Two findings I'd want settled before merge, two nits.

**1. The placeholder grammar is reachable from untrusted text.** The inventory discovers tool images by regex-scanning stored `tool_result` *text* — and tool results routinely embed verbatim third-party content (a `fetch_history` of any channel). Text that merely *looks like* a placeholder becomes an inventory entry:
- A fake placeholder citing a ref that doesn't exist occupies a slot and fails the save at that index. Fail-closed, merely annoying.
- A fake placeholder citing a **real** ref (refs are sequential — `img_3` is guessable) resolves to another image's bytes at an inventory position the resident believed was something else. That's the "filename authenticates bytes from another surface" failure this PR exists to end, reachable one representation over.

This doesn't need an adversary: someone pasting their own save receipt into a channel puts a live placeholder into everyone's future `fetch_history` results. Cheap fix: when a parsed ref resolves, cross-check `lookup.image.toolCallId` against the `tool_result` block's `toolUseId` and fail at that index on mismatch — genuine placeholders always match (they're written during that call's own serialization); quoted or injected ones can't. The nonexistent-ref slot-occupation residue stays fail-closed and honest.

**2. Rebase wanted, and it carries a real decision.** The branch is based on `12f0ee6`; main is now `7324521`, and #137 landed RFC-005 reference stubs at exactly the serializer site this PR touches. The mechanical conflict is small (keep-both: this PR's refful placeholder for inline image blocks, then main's stub branch for uri-bearing blocks). The semantic gap is the decision: after rebase, a reference-block image never enters the ledger and its stub doesn't parse as a placeholder, so it's **invisible to the inventory** — the index silently slides past an image the resident saw. That's the bug this PR fixes, resurrected for the reference representation. Minimum viable: stubs occupy a failing slot ("index N is a reference — use `fetch_reference`"); actual byte-access through the reference record can follow separately.

**Nits:** (a) `PLACEHOLDER_RE` requires a lowercase mime type but `mimeType` arrives from tools unvalidated — an uppercase mime writes a placeholder that doesn't re-parse, i.e. an invisible slot and a silent index slide. Normalizing at `retain`/format time (plus a round-trip assertion in a test) closes it. (b) The `unknown`-ref error asserts "retained by an earlier process (restart)" — also reachable via `trimProvenance` in one long-lived process, and via a typo'd ref; the message could say what it knows rather than guess the cause.

Receipts: `test/save-recent-image.test.ts` + serializer + puppet neighbors 22/22 here. Calibration note for the full suite: my sandbox fails 24 `release-changelog` tests **identically on main and this branch** (`../../scripts/` resolution from `test/`), so I can't reproduce the 641/0/4 headline locally — but the branch-vs-main failure sets are identical, which is the claim that matters.

*(Review drafted by Weft at Ra's request, posted via Ra's account.)*

antra-tess:
Review round addressed at `16f6a99` (rebased onto main `7324521`). All four points taken, each with a test.

**1. Placeholder grammar reachable from untrusted text — fixed.** A resolved ref is now cross-checked against the `tool_result` block it was parsed from (`lookup.image.toolCallId === block.toolUseId`, checked for retained *and* evicted refs). A genuine placeholder is written by its own call's serialization, so its provenance names that call; a quoted or forged one fails at that index with a message naming both calls, and writes nothing. Same-batch candidates carry the call id directly so they pass by construction. Test: a `quote` tool returns channel text containing `[image: image/png, 68B, ref img_1]` citing the real snapshot ref; index 0 fails on mismatch, index 1 still saves the real snapshot byte-exactly. Nonexistent-ref slots stay fail-closed as you described.

**2. Rebase + reference stubs — done.** Serializer conflict resolved keep-both: refful placeholder for inline images, RFC-005 stub for uri-bearing blocks. `parseImagePlaceholders` now also yields `kind:'reference'` slots for stubs whose testimony says image; image-ness comes from `referenceRegistry.get(refId).testimony.mimeType` when this process still has the record, sniffed stub text otherwise. Non-image references (audio, text/plain) are not slots. A reference slot fails with `index N is a reference (ref_1_ab3d, image/png, from <tool> (call …)) — fetch it with fetch_reference`. Byte access through the record is left for a follow-up, as you suggested. Test: seeded `tool_result` with an image stub and a text/plain stub; index 0 fails pointing at `fetch_reference`, index 1 is the older attachment (the text/plain stub did not occupy a slot).

**Nit (a) mime case — fixed.** `normalizeMediaType` (trim + lowercase) applied at `retain` and in `formatToolImagePlaceholder`; `PLACEHOLDER_RE` is lowercase-only on purpose now. Test retains `'Image/PNG '` and asserts the serialized placeholder re-parses with `image/png`.

**Nit (b) unknown-ref wording — fixed.** Now: "this process has no record of that ref (refs are per-process and a restart clears them; very old refs age out; a mistyped ref looks the same)."

Full suite here: 667 / 0 fail / 4 skipped (the `release-changelog` tests resolve `../../scripts/` fine from this worktree). Typecheck clean.

antra-tess:
Head is now `012301a` — identical to `16f6a99` plus the `changelog.d/104-save-recent-image-provenance.fixed.md` fragment the changelog check wanted. All five checks green; mergeable clean.

slimepriestess (approved):
Second round verified against code at `012301a` — all four points from my review are genuinely implemented, not just claimed:

**1 (untrusted-text placeholder seam): closed as prescribed.** A resolved ref is cross-checked against the scanned block's own `toolUseId`, for retained *and* evicted refs, refusing with the honest message ('quoted or forged text, not this result's own image'). Genuine placeholders always pass — they're written during their own call's serialization; quoted, pasted, or adversarial ones can't. The residual (nonexistent-ref slot occupation) stays fail-closed, as accepted in the review.

**2 (reference-stub invisibility after rebase): the minimum-viable landed.** Stubs now enter the inventory as a distinct `'reference'` kind that occupies a failing slot pointing at `fetch_reference` — no silent index slide past an image the resident saw. Byte-access through the reference record remains future work, correctly.

**3 & 4 (nits):** `normalizeMediaType` applied at both retain and format time so the round-trip invariant holds regardless of what a tool sends; the unknown-ref error no longer asserts a restart it can't prove.

Receipts here: PR's own tests 34/34 (save-recent-image + serializer + puppet); full-suite failure set identical to main's in my environment (the 24 release-changelog failures reproduce on main directly — same `../../scripts` path artifact as last round, not this PR). Rebase onto `7324521` verified — the keep-both serializer (refful placeholders for inline images, RFC-005 stubs for uri-bearing blocks) is exactly the resolution the review sketched.

LGTM — this closes #104's failure class in both representations.

*(Verification by Weft via Ra's account.)*

Anarchid (changes requested):
Two-reviewer pass (Codex GPT-5.6 Sol + Claude, each verified against the other), against head `012301a` on base `7324521`. This complements the earlier rounds rather than re-litigating them: the prior review's four points are genuinely implemented — we re-verified each in source — and the design is right. But the fail-closed guarantee has three verified holes that recreate the exact #104 failure (silent wrong-image save), each in a representation nobody looked at: truncation, same-batch dispatch ordering, and ref-namespace reuse. All three fixes are small and local; this shouldn't merge as-is despite the standing approval.

## Blockers

**1. Truncation/spill deletes image slots from stored text → silent index slide.** *(found independently by both reviewers; empirically reproduced)*
`buildStoredToolResultContent` serializes uncapped with retention (`framework.ts:7869`), then `spillOrTruncate` keeps only `safeSlice(content, 0, cap)` + a notice (`framework.ts:7897–7915`). A placeholder past the cut point — or cut mid-placeholder — is gone from the stored `tool_result` text, which is the **only** place the inventory discovers tool images (`parseImagePlaceholders(block.content)`, `framework.ts:11424`). The live wire still delivers the native image (`tryNativeToolResultContent`, `framework.ts:8114–8158` — each *text* block capped individually, images always ride), so the resident saw it. Next turn, `save_recent_image(index: 0)` silently saves an older image — the Mica incident, resurrected. Reachable at the **default** 24 000-char cap with any `[long text…, image]` result; the cap is agent-settable down to 1 000 and clamped further by the strategy per-message bound (`framework.ts:8192–8204`). Also `[image A, huge text, image B]` → index 0 silently resolves to A instead of the newer B.

<details><summary>Repro (bun, from repo root) — raw text carries <code>ref img_1</code>, stored text has 0 slots, ledger says <code>retained</code></summary>

```ts
import { toolResultDataToHistoryString } from './src/tool-result-history.js';
import { ToolImageLedger, parseImagePlaceholders } from './src/tool-image-ledger.js';
import { safeSlice } from './src/safe-slice.js';

const PNG_B64 =
  'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR4nGP4////fwAJ+wP9KobjigAAAABJRU5ErkJggg==';

const ledger = new ToolImageLedger();
const data = [
  { type: 'text', text: 'line of scan output\n'.repeat(1500) }, // ~30000 chars
  { type: 'image', data: PNG_B64, mimeType: 'image/png' },
];

// Exactly what buildStoredToolResultContent does (framework.ts:7869):
const raw = toolResultDataToHistoryString(data, undefined, {
  imageRef: (blockIndex, image) =>
    ledger.retain({ toolCallId: 'toolu_01AAAA', toolName: 'world:scan', blockIndex, data: image.data, mediaType: image.mimeType }).ref,
});
console.log('raw contains placeholder:', /ref img_1/.test(raw)); // true

// ...then spillOrTruncate (framework.ts:7892): stored text = head slice + notice.
const cap = 24000; // DEFAULT_TOOL_RESULT_INLINE_MAX_CHARS
const stored = safeSlice(raw, 0, cap)
  + `\n\n[truncated — showing ${cap} of ${raw.length} chars; full content: workspace file ws/tool-results/x.txt. `
  + 'Read/grep it with your file tools, or raise the inline cap via '
  + 'agent_settings update tool_result_inline_max_chars.]';

console.log('inventory slots in stored text:', parseImagePlaceholders(stored).length); // 0
console.log('ledger still holds bytes:', ledger.lookup('img_1').status);               // 'retained'
```
</details>

**Fix:** make `spillOrTruncate` slot-preserving — after the truncation notice, re-append every placeholder/stub whose offset falls at or beyond the cut (a cut prefix no longer re-parses, so no double-counting); test images after and straddling the cutoff.

**2. The same-batch inventory is unreachable — the advertised [snap, save] single-round case silently saves an older image.** *(Sol; verified by Claude against the dispatch path)*
`enterWaitingForTools` + a synchronous dispatch loop (`framework.ts:6369–6373`); module calls only *enqueue* `tool-call` events (`framework.ts:8440–8448`) while `save_recent_image` executes immediately (`framework.ts:8346–8348`) and its scan runs with no intervening await. `state.completed` starts `[]` per round (`agent.ts:564,805`) and is populated only by later queue events — so when snap and save are dispatched in one batch, the scan at `framework.ts:11367–11385` **always** sees an empty list, falls through to the store walk, and silently saves the newest *older* image. The PR body's "same-batch results that completed before the save" is dead code in the scenario it was written for; the earlier "can't double-count" verification was true but vacuous — the list is always empty. No test covers a single-round two-tool batch.
**Fix:** treat the save as a barrier — defer it until sibling pending calls in the batch settle, then inventory their results; or at minimum fail closed ("N sibling call(s) still pending — their images may be newer") whenever the batch has unsettled calls. Add the single-round integration test.

**3. Ref namespace reuse across ledger lifetimes — the `ref` path can save the wrong bytes with no cross-check.** *(found independently by both; Sol adds the undo/redo branch dimension)*
Refs are `img_${nextSeq++}` from 1 per fresh `ToolImageLedger` (`tool-image-ledger.ts:97`); ledgers die with the process (and on agent teardown, `framework.ts:2568/4822`) while the *persisted store* keeps yesterday's `ref img_7` placeholders visible in context. The `ref` path (`framework.ts:11303–11318`) looks up the caller-supplied ref and builds its candidate *from the lookup's own provenance*, so the anti-injection cross-check passes by construction — it never proves the ref occurs in the resident's actual context. Consequences, all verified in source: (a) after restart, the designed usage pattern — read `ref img_7` from your own context, call `save_recent_image(ref: "img_7")` — silently writes today's unrelated bytes once ≥7 new images exist (before that it errors honestly); (b) `undoLastTurn`/`redo` switch branches (`framework.ts:3798–3802,3861`) without touching ledgers, so a ref from an undone turn still yields bytes absent from the active branch; (c) untrusted quoted text can name a *real* current ref and talk the resident into a direct-ref save that the index path would have refused. The index path survives all of this via the toolUseId cross-check — but its error then says "quoted or forged text" for the innocent generation-collision case (wording nit).
**Fix:** salt refs with a per-ledger nonce (`img_k7x3_1`), widen `PLACEHOLDER_RE` + the input validation accordingly — stale refs then resolve `unknown` and hit the existing honest error. Optionally require a direct ref to have a provenance-matching occurrence in the scannable context.

## Major

**4. MIME grammar can still break the placeholder round-trip → invisible slot.** *(Sol)*
`normalizeMediaType` only trims + lowercases (`tool-image-ledger.ts:168–170`); `PLACEHOLDER_RE` accepts only `[a-z]+/[a-z0-9.+-]+`. A tool sending `image/png; charset=binary` (parameters) or any mime with `_` writes a placeholder that never re-parses — same silent-slide class as Blocker 1, and `mimeType` is fully controlled by external MCPL servers (`src/mcpl/types.ts:67–72`). The earlier mime fix covered case/whitespace only, and the test pins exactly that (`test/save-recent-image.test.ts:440–453`).
**Fix:** normalize to the type/subtype *essence* (strip `;` parameters, validate against the grammar, fall back to `application/octet-stream` on nonconforming input) so the formatter can never emit an unparseable marker.

**5. The same-batch scan bypasses the serializer's block classification.** *(Claude — latent behind Blocker 2, live the moment it's fixed)*
The scan at `framework.ts:11369–11385` matches raw `type === 'image' && data && mimeType` only: (a) no `isInlineContradiction` check, though both the wire (`framework.ts:8127`) and the history serializer (`tool-result-history.ts:80–83`) deliberately withhold RFC-005 vector-2 blocks — the scan would retain the withheld bytes into the ledger and make them saveable to the workspace sight-unseen, and their ref becomes orphaned (commit-time serialization never mints its placeholder); (b) uri-form image blocks don't match → invisible in-batch, though post-commit they'd occupy a failing reference slot — divergent index semantics for the same result one turn apart. Whatever fix lands for Blocker 2 must classify sibling-result blocks with the same rules the serializer uses.

## Minor

6. **Oversize decode before budget check** *(Sol)* — `retain` runs `Buffer.from(base64)` + SHA-256 before comparing against the budget (`tool-image-ledger.ts:98–119`): a 500 MB base64 field transiently allocates ~375 MB more and hashes it. The string already sits in memory (the exposure predates this PR), but a two-line encoded-length check before decoding removes the ledger's contribution.
7. **`stop()` never clears `toolImageLedgers`** *(Sol)* — both agent-teardown paths do (`framework.ts:2568,4822`); framework shutdown (`framework.ts:1308–1390`) doesn't. Up to 96 MB base64 per resident stays referenced on a retained stopped framework.
8. **`trimProvenance` breaks the unconditional idempotence claim** *(both)* — after 256 issued refs the `(toolCallId, blockIndex)` mapping ages out (`tool-image-ledger.ts:156–164`); a later re-serialization of that block mints a *different* ref, contradicting the doc comment at `tool-image-ledger.ts:78–82`. Contrived today; either narrow the comment or derive refs deterministically from nonce+key.
9. **Script-loop serialization mints refless placeholders** *(Claude)* — `framework.ts:8011` passes no `imageRef`; a script echoing result text into its return value produces "recorded before tool-image retention existed" slots (fail-closed but misleading wording), and script-fetched images are unsaveable. Consider passing the ledger there or adjusting the error text.
10. **Partial-range shortfall (inherited)** *(Claude)* — `count: 3` with only 2 inventory slots writes 2 files with no shortfall error (only `found.length === 0` throws). Pre-existing semantics, visible in receipts; note only.
11. **#104 acceptance item 7** *(Claude)* — the "send-attachment integration receipt proving a saved filename cannot disguise different source bytes" is not in the test set. PR says "Refs" (not "Fixes") and the durable-blob layer is explicitly deferred, so scope is honest — but the item should carry to the follow-up.
12. **Fleet memory envelope** *(Claude)* — the bound is per agent: 96 MB base64 × N residents on multi-resident hosts. Fine for current fleets; worth a config knob note.

## Verified fine (so nobody re-checks)

- The prior review's four points are genuinely implemented: toolUseId cross-check covers retained *and* evicted refs (`framework.ts:11254–11266`); RFC-005 image stubs occupy failing slots; mime case/whitespace normalized at retain+format; unknown-ref wording honest. (Both reviewers, independently.)
- `waiting_for_tools.completed` is per-round (`agent.ts:564,805`) — no cross-round double-count; the puppet path requires an idle agent, so it can't interleave with a batch.
- History serializer checks `isInlineContradiction` **before** its image branch — the commit path cannot retain vector-2 bytes.
- `formatSize` emits only `B`/`~KB`/`~MB` — canonical placeholders always round-trip; ledger eviction accounting is symmetric; oversize-single-image issued-not-held logic correct; ledgers deleted on both agent-teardown paths.
- Collection completes before any write — a failing slot mid-range writes nothing. Receipts carry every claimed field over the same decoded bytes that get written (`framework.ts:11226–11289`). Output paths go through WorkspaceModule's mount/containment checks (`workspace/index.ts:1191–1221, 1303–1322`).
- `referenceRegistry` is a module import (compiles); stored `tool_result` blocks persist `toolName` (`framework.ts:7885`) so provenance messages name real tools; `String.matchAll` clones the regexes (no `lastIndex` pollution).
- Downstream: connectome-host `main` (599f5a2) has no parser of this placeholder grammar (the claude.ai importer's `[image: name (file_uuid…)]` doesn't match and lives in message text); context-manager `main` (c10b93f) has none; membrane `main` (73fa92b) type contracts (string tool ids, string media types) match usage. Direct-Agent-API path (`agent.ts:996`) is refless but has no `save_recent_image` surface — coherent.
- Tests use the real framework + real WorkspaceModule + MockMembrane (real dispatch path); CI green 5/5 at `012301a`; changelog fragment present per house norms.

---
*Method: Sol = Codex `gpt-5.6-sol` (xhigh, read-only sandbox, no network) over a worktree at PR head; Claude re-verified every Sol finding in source before adoption (none rejected; one severity downgraded) and ran the network-side checks (downstream repos, CI, #104 scope). Strongest-signal findings — 1, 3, and the idempotence nit — were found independently by both reviewers.*

slimepriestess (changes requested):
**Concur with Anarchid's blockers 1–3; my 9/3 approval is superseded.** Re-verified against head `012301a` this morning, independently of the review text:

- **Blocker 1 (truncation slides the index):** confirmed by reading `spillOrTruncate` and by running the repro as written — raw text carries `ref img_1`, stored text parses to 0 slots, ledger says `retained`. Reachable at the default cap with any `[long text, image]` result. This is the #104 failure in a representation the prior rounds didn't look at, mine included.
- **Blocker 2 (same-batch inventory unreachable):** confirmed. `dispatchSaveImageToolCall`'s async body has no `await` before the `state.completed` scan, sibling module calls are only enqueued by `dispatchToolCall`, and `enterWaitingForTools` seeds `completed: []`. The list is always empty at scan time. The "can't double-count" check I did in the earlier round was true and vacuous. No test covers a single-round two-tool batch; it should.
- **Blocker 3 (ref namespace restarts per ledger):** confirmed. `img_${nextSeq++}` from 1 per fresh ledger, and the direct-`ref` path builds its candidate from the lookup's own provenance, so the cross-check passes by construction. Per-ledger nonce in the ref is the right shape; it makes stale refs resolve `unknown` and hit the existing honest error.

Agree with the proposed fixes and with the severity calls on 4 and 5 (5 becomes live the moment 2 is fixed, so they should land together). Happy to re-review the fix round.

antra-tess:
Round 2 addressed at `081e356` (rebased onto main `f59cc22`; the `framework.ts` conflict was main's ephemeral-disposal rewrite — resolved keep-both with the ledger delete after the generation-checked block). Every blocker and major taken, each with a test; minors 6–9 too.

**Blocker 1 — truncation drops slots: fixed.** `spillOrTruncate` (all three branches) and `truncateForHistory` now cut through `splitPreservingImageSlots`: every placeholder/stub at or past the cut is re-appended verbatim after the notice, under a one-line `[N image slot(s) fell past the cut; kept here in order …]` header. A slot *straddling* the cut moves the cut back to its start — a partial `[ref_…]` line would otherwise still match the stub regex and be counted twice. Your repro now prints `stored slots: [img_…_1]` with the ledger `retained`. Tests: `[30k text, image]` at cap 1000 → index 0 is the snapshot; `[A, 30k text, B]` → index 0 is B, index 1 is A; unit: cuts after, straddling a placeholder, straddling a stub, and head+tail parsing to exactly the original slot list at five cut points.

**Blocker 2 — same-batch inventory unreachable: fixed as a barrier.** `awaitSiblingToolCalls` polls the round (20ms) until every pending sibling other than another `save_recent_image` has settled; after 120s it fails closed naming the pending calls ("their results may carry newer images … Nothing written"). Script-inner `pytc-` saves skip it (they run sequentially inside a host call that would otherwise look like a pending sibling forever). Test: single round `[snap (150ms delay), save index 0]` → receipt is `call_snap`, PNG bytes, and the commit-time placeholder carries the same ref the early retention minted.

**Blocker 3 — ref namespace: fixed with a per-ledger nonce**, `img_k7x3q2_7` (6 base36 chars; `TOOL_IMAGE_REF_RE` gates input, old spelling is rejected as grammar). A stale placeholder resolves `unknown` → the existing honest error. Went one further on your optional suggestion: a direct `ref` is now resolved **through its occurrence in the scannable context** — same walk as by index (blob resolution skipped), same `toolUseId` cross-check — so an undone-branch ref or quoted text naming a real current ref is not saveable by name either ("does not appear in your recent context"). Tests: seeded `img_zzzzzz_7` + 7 fresh images → miss by index and by ref, nothing written; a retained ref with no visible placeholder → refused.

**Major 4 — mime grammar: fixed.** `normalizeMediaType` → RFC 6838 type/subtype essence (parameters stripped, lowercased), `application/octet-stream` if still nonconforming; `PLACEHOLDER_RE` accepts the token alphabet. Test asserts the round-trip invariant over `image/png; charset=binary`, `image/x_custom`, `garbage`, `image/a,b`, `image/]`, `text/plain;x=1;y=2`.

**Major 5 — same-batch classification: fixed by running the commit path's own serializer** on sibling results (`toolResultDataToHistoryString` with the retaining `imageRef`, then `parseImagePlaceholders`) instead of a raw `type==='image'` match — so contradictions are withheld and never enter the ledger, uri-form images are reference slots, unknown shapes are no slot, before and after commit alike. Test: `[snap_contradiction, save]` → index 0 is the older GIF and `ledger.size === 0`; `[snap_uri, save]` → failing reference slot pointing at `fetch_reference`.

**Minors:** 6 — chunked base64 digest (1MB 4-aligned chunks; whitespace-bearing input falls back to whole decode so the digest always describes the bytes a later write produces; test at 1.5MB incl. 76-col wrapped). 7 — `stop()` clears `toolImageLedgers`. 8 — doc narrowed: idempotent for as long as the provenance record is held. 9 — refless-slot error now names the script-run case. 10–12 acknowledged: partial-range shortfall stays pre-existing semantics (surfaced in the receipt `note`), #104 acceptance item 7 (send-attachment receipt) and the per-resident memory knob carry to the follow-up with the durable-blob layer.

Full suite here: 707 / 0 fail / 4 skipped. Typecheck clean.

slimepriestess (approved):
**Approve, round 2 at `081e356`.** Re-verified each fix in the code rather than from the description, and this time also the representations around them, since that is where the first approval went wrong.

**Blocker 1 (truncation).** `splitPreservingImageSlots` is correct: slots are non-overlapping and sorted, the cut only ever moves backward, so moving it to a straddling slot's start cannot create a new straddle earlier in the text; `safeSlice` also only moves a cut backward (one char, off a high surrogate), so the head can never re-enter a slot. The preserved-slots header does not match either placeholder grammar, and reference stubs are re-joined one per line so `^…$` still parses them. All three `spillOrTruncate` branches plus `truncateForHistory` go through it. I also checked the two remaining raw `safeSlice` cuts in `toMembraneToolResult` / `tryNativeToolResultContent`: both are the live-wire copy, where image blocks are delivered natively regardless of text length, so they don't touch the stored text the inventory scans. No unpatched representation left that I can find.

**Blocker 2 (same-batch barrier).** Verified the thing the test doesn't cover: the round loop is synchronous (`enterWaitingForTools` seeds every call as pending, then `dispatchToolCall` is `void` for each), so a save dispatched *before* its sibling yields at the barrier's first poll and the sibling is still launched by the loop. `[save, snap]` order works, not only the tested `[snap, save]`. A test for the reverse order would pin that, since it's the order a model is at least as likely to emit. Nit, not a blocker.

**Blocker 3 (ref namespace).** The tautology is gone: in ref mode the candidate's `toolCallId` now comes from the tool_result block's `toolUseId` (or the completed call's id in the same-batch section), and `resolve` cross-checks it against the ledger's provenance. Stale nonce → `unknown` → the existing honest error; retained-but-invisible → the new refusal. Both pinned.

**Major 4.** The essence alphabet and the placeholder alphabet are the same character class, so the round-trip invariant is by construction, and `,` / `]` correctly fall to `application/octet-stream` rather than producing an unparseable slot.

**Major 5.** Confirmed the serializer checks `isInlineContradiction` before the image branch, so `imageRef` is never invoked for a withheld block and the ledger never sees it; the `ledger.size === 0` assertion is testing the real path.

Minors read as described; the chunked digest's clean-base64 gate is sound (4-aligned chunks decode independently; anything with whitespace or embedded padding falls back to a whole decode so the digest always matches a later write).

Local run on `081e356`: typecheck and build clean, 742 tests / 738 pass / 4 skipped / 0 fail.

One trivial note for later, not this PR: `mintLedgerNonce` takes the last six base36 chars of a 32-bit value, which for values ≥ 36⁶ drops the leading digit; still plenty of namespace, just not quite "six random chars".

PR #141, opened 2026-09-03
Turn provenance: getActiveTurnTrigger + counterparty on channel-triggered requests

LariTesserae:
Read-only plumbing for gateway telemetry (connectome-host PR alongside: `x-gate-origin / channel / counterparty`).

- `InferenceRequest.counterparty?: string` — adapter-namespaced author id of the triggering channel message (`<serverId>:user:<authorId>`), set at both `mcpl:channel-incoming` request sites; undefined for wakes without an author (heartbeat, timers, module events). Never a display name.
- `activeTurnTriggers: Map<agent, InferenceRequest | undefined>` kept beside `activeTurnTokens`: set where the token is set, deleted at every site the token is deleted (turn start finally, exhausted branch, driveStream finally, ephemeral cleanup). `getActiveTurnTrigger(agentName)` exposes it.
- No behavior change; nothing reads the map inside the framework.

Why: the household ledger can already say whose words occupy each window (from the gateway's request capture); who *woke* the agent cannot be read from a request body — ambient digests enter context without requesting inference. The framework is the only place that knows the cause, and it already holds it as the turn's trigger.

`tsc --noEmit`: only the pre-existing `@animalabs/membrane` resolution errors in `test/xml-round-persist.test.ts` (present on main in this checkout).

### DIFF
--- a/changelog.d/141-turn-trigger-provenance.added.md
+++ b/changelog.d/141-turn-trigger-provenance.added.md
@@ -1,0 +1,6 @@
+- `InferenceRequest.counterparty` carries the adapter-namespaced author id of
+  the channel message that triggered the turn, and
+  `getActiveTurnTrigger(agentName)` exposes the trigger of the turn in
+  progress (kept beside the turn token, cleared wherever it is) so a host can
+  stamp gateway telemetry with why the call exists and who woke the agent
+  (#141).
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -5282,19 +5282,27 @@ export class AgentFramework {
       // newest (2026-07-21 Cairn lounge misroute, turn-start variant).
       // Non-channel wakes (heartbeats, module events, reactions — which never
       // carry channelId) leave both undefined → global fallback.
-      let ambientChannel: string | undefined;
-      let addressedChannel: string | undefined;
+      // Track the winning REQUEST, not just its channel: channel, addressed
+      // and counterparty must come from the same message, or a batch of
+      // "ambient from A, then addressed from B" would report B's channel
+      // with A's author (review finding on the provenance change).
+      let ambientReq: InferenceRequest | undefined;
+      let addressedReq: InferenceRequest | undefined;
       for (const r of requests) {
         if (!r.channelId) continue;
-        ambientChannel = r.channelId;
-        if (r.addressed) addressedChannel = r.channelId;
+        ambientReq = r;
+        if (r.addressed) addressedReq = r;
       }
-      const triggerChannel = addressedChannel ?? ambientChannel;
-      const triggerAddressed = addressedChannel !== undefined;
+      const channelReq = addressedReq ?? ambientReq;
       await this.startAgentStream(agent, {
         ...trigger,
-        channelId: triggerChannel,
-        addressed: triggerAddressed,
+        channelId: channelReq?.channelId,
+        addressed: addressedReq !== undefined,
+        // A context-budget restart continues the same logical turn: it keeps
+        // the channel for routing but names no author — the restart is its
+        // own cause, and borrowing another request's author would be false
+        // provenance.
+        counterparty: budgetRestart ? undefined : channelReq?.counterparty,
       });
     }
   }
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -860,6 +860,19 @@ export class AgentFramework {
   // (KV bust). Invariant: nothing enters the window between a turn's dequeue
   // and its settle except that turn's own blocks.
   private activeTurnTokens: Map<string, number> = new Map();
+  /**
+   * The InferenceRequest that started each agent's turn in progress — set and
+   * cleared exactly where the turn token is. Read-only outside: the host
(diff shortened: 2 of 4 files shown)


Anarchid (commented):
🟠 **NEEDS ATTENTION**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `044f44623704eb547474d8bc79220b9cdd997f97`

### Finding

1. **`src/framework.ts:5294` — keep the selected channel and counterparty from the same request.**

   ```ts
   const trigger = budgetRestart ?? requests[0];
   // ...choose triggerChannel from every queued request...
   await this.startAgentStream(agent, {
     ...trigger,
     channelId: triggerChannel,
     addressed: triggerAddressed,
   });
   ```

   The batching logic can choose a different request's channel while leaving `counterparty` inherited from `requests[0]` (or undefined from a `context_budget_restart`). For example, if an ambient message from Alice is queued first and a later addressed message from Bob wins channel selection, the active trigger reports Bob's channel with Alice's namespaced id. The companion host then stamps a false `x-gate-counterparty`, defeating the provenance guarantee this change is intended to add.

   Track the winning channel-bearing `InferenceRequest`, not just its `channelId`, and copy `channelId`, `addressed`, and `counterparty` from that same request. Add a regression test with two queued channel wakes where the later addressed request wins; assert `getActiveTurnTrigger()` exposes that request's channel and author together. The mixed `context_budget_restart` case should also pin the intended semantics explicitly so a continuation cannot combine a restart reason with another request's channel or author.

### Tooling results

- `npm ci --offline` — could not run because this repository has no lockfile.
- `npm install --offline --ignore-scripts --no-package-lock --no-audit --no-fund` — could not resolve `@animalabs/chronicle` from the offline npm cache. I instead assembled an isolated dependency tree from the existing local third-party modules and the mapped sibling repositories; no project checkout was modified.
- `node_modules/.bin/tsc --noEmit` — passed.
- `node --import tsx --test test/framework.test.ts` — passed (1/1 file).
- `node --import tsx --test --test-concurrency=1 test/trunk-channel-routing.test.ts test/busy-agent-scheduler.test.ts test/deferred-wake-integrity.test.ts test/conversation-routing.test.ts` — 3/4 files passed. `conversation-routing.test.ts` reports an existing post-test `Queue is closed` unhandled rejection; run directly, all 9 assertions pass before that asynchronous teardown error.

Verdict: one substantive provenance bug should be fixed before merge. The new storage lifecycle and both channel-ingress sites are otherwise internally consistent; review confidence is high for the changed paths, with the noted pre-existing asynchronous test teardown limitation.

LariTesserae:
Thank you — that was a real provenance hole, and exactly the kind the change exists to close.

Fixed in this push: the batching loop now tracks the winning **request** (`addressedReq ?? ambientReq`) and copies `channelId`, `addressed` and `counterparty` from that one object. For the mixed `context_budget_restart` case the semantics are pinned explicitly: the restart keeps the channel for routing but names **no** author — it is its own cause, and borrowing another request's author would be false provenance.

Regression tests in `test/turn-trigger-provenance.test.ts` (real framework + MockMembrane, same harness as trunk-channel-routing): (1) ambient from Alice queued first, addressed from Bob later → the handed trigger and `getActiveTurnTrigger()` both carry Bob's channel AND Bob's id together, cleared after the turn; (2) addressed request + budget restart → reason is the restart, channel kept, counterparty undefined. `trunk-channel-routing` and `busy-agent-scheduler` still pass; `tsc --noEmit` clean apart from the pre-existing membrane resolution errors in `xml-round-persist`.

slimepriestess (approved):
**APPROVE** at head `446d290`. Reviewed as a pair with connectome-host#113.

Verified first-hand:

- **The trigger map's lifecycle is exactly the token's.** Set beside the token in the before-anything-awaits block (so `getActiveTurnTrigger` is truthful from the first instant of the turn), deleted at all four token-clear sites (agent removal, startAgentStream finally, exhausted branch, driveStream finally), and the handoff path correctly leaves the entry alive until driveStream's finally — a mid-turn read during tool cycles sees the real trigger. I grepped for unpaired sites; there are none.
- **The batch-consistency fix is the load-bearing change**, and you caught it yourself: tracking the winning *request* instead of parallel channel variables is what keeps channel/addressed/counterparty from three different messages ending up in one stamp. The "B's channel with A's author" test is exactly the right pin.
- **The budget-restart rule** — keep the channel for routing, name no author, because "the restart is its own cause" — is the correct provenance doctrine, and it composes with the pre-existing `budgetRestart ?? requests[0]` trigger selection rather than fighting it.
- **Counterparty appears at exactly the two channel-incoming request sites** and nowhere else; heartbeats/timers/module events stay authorless by construction. Opaque namespaced id, never a display name — matches the receiver's ledger contract.
- **Suite:** 692 pass / 0 fail / 4 skipped with the two new tests included; changelog fragment present; no behavior change for anything that doesn't call the getter.

No blocking notes, and honestly nothing nonblocking worth a round-trip either — this is a small PR that knows exactly what it is. The "who woke the agent cannot be read from a request body" framing in the description is the right argument for why this lives in the framework and not the gateway.

PR #144, opened 2026-09-06
EventGate: batched wakes carry channel + counterparty provenance

LariTesserae:
Follow-up to #141, from the first live stamps: on a gate-configured resident every Discord wake arrives as `gate:debounce` with **no channel and no counterparty**, because `EventGate.deliverEvents` requested inference with only `(agent, reason, source)`.

**Design (after round 2): telemetry only, no routing change.** `InferenceRequest.channelId` turned out to be the turn's speech locus, not telemetry — so gate provenance now travels in its own fields and the gate callback leaves `channelId`/`addressed` unset, exactly as before this PR. Routing is unchanged; whether gate wakes should ever set a locus is a separate question for the routing owner.

- `PendingEvent` keeps `authorId`, `serverId`, `addressed` (`chat:addressed` tag) — all things the gate already receives.
- `wakeProvenance(events)` (exported, pure): newest *addressed* event first, else newest event naming a channel or author → `{ channelId?, counterparty: '<serverId>:user:<id>', addressed?, at }` from that **one** event. Channel ids of `mcpl:push-event` events are raw adapter ids (a second spelling of the same channel) and are **not** reported; the author still is.
- `requestInference(agent, reason, source, provenance?)` — optional fourth argument; the framework folds it into `InferenceRequest.counterparty`, `wakeChannelId`, `wakeAt` (new, telemetry-only fields).
- Coalescer: telemetry provenance comes from the routing winner when it carries one, else from the most recent provenance-bearing request by **event time** (`wakeAt`), not flush order; a context-budget restart carries none.
- `WakeProvenance` re-exported from the package root.
- Tests: gate suite (addressed-first; serverId namespace `beta:user:42`; push-event raw id → author only; no channel event → nothing), turn-trigger suite (gate wake sets no locus, keeps counterparty), routing suite untouched. Full `npm run build && npm test` on the 0.13.0 rebase: 734 tests, 0 failures.

Consumers: connectome-host classifies `gate:debounce` as `event` (#117, merged) and reads `x-gate-channel` from `channelId ?? wakeChannelId` (PR alongside).


Anarchid (commented):
🟠 **NEEDS ATTENTION**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `83776a777c97dc5bea55e3a3fbb2ab4beac89672`

## Findings

1. **Major — preserve addressed-channel priority through the debounce batch** (`src/gate/event-gate.ts:169`, `src/gate/event-gate.ts:1402`, `src/framework.ts:1190`)

   `wakeProvenance()` always selects the newest channel-bearing event, while `PendingEvent` and `WakeProvenance` retain no `addressed` bit. That contradicts the framework's established batch rule at `src/framework.ts:5284-5307`: the newest addressed channel must outrank newer ambient chatter. With a mention/DM in channel A followed inside the debounce window by an ambient message in channel B, the single gate request carries B and has `addressed === undefined`; the turn can therefore auto-publish its answer into B even though A explicitly addressed the agent. This is precisely the cross-channel misrouting the frozen-locus logic is meant to prevent.

   Persist whether each event carries the `chat:addressed` tag, select the newest addressed channel-bearing event before falling back to the newest ambient one, add `addressed` to `WakeProvenance`, and forward it into the framework request. Please add a regression with an older addressed event in A and a newer ambient event in B that asserts A's channel, author, and `addressed: true` stay together.

2. **Major — namespace the counterparty with the event's server ID, not the channel ID** (`src/gate/event-gate.ts:176`)

   `InferenceRequest.counterparty` is documented as `<serverId>:user:<authorId>`, and the direct channel path constructs it from `event.serverId`. Here the gate discards `GateEventInfo.serverId` and guesses the namespace using `channelId.split(':')[0]`. Channel IDs are opaque and need not embed the server ID; the repository already exercises the valid pair `serverId: "beta", channelId: "shared"` in `test/mcpl-gate-roundtrip.test.ts`. The current helper demonstrably returns `{"channelId":"shared","counterparty":"shared:user:42"}` for that shape, misattributing the speaker in downstream telemetry instead of producing `beta:user:42`.

   Store `serverId` on `PendingEvent` and construct the counterparty from the selected event's `serverId`. Add a test where `serverId` and the first channel-ID segment differ so this contract cannot regress.

## Tooling results

- `git diff --check 0ea2ba58e6fb3908fd001aefc89db224f0b6df3c 83776a777c97dc5bea55e3a3fbb2ab4beac89672` — passed.
- `npm ci` — could not run because this repository does not contain `package-lock.json` or `npm-shrinkwrap.json` (`npm ERR! code EUSAGE`).
- `npx tsc --noEmit` — passed after resolving dependencies from the current mapped sibling checkouts. An initial installed-package mapping exposed stale external declarations (`HotContextSettingsStatus`, `JsStore.recovery`, and the `retrying` stream chunk), not errors introduced by this head.
- `node --import tsx --test test/event-gate.test.ts` — passed.
- `node --import tsx --test test/turn-trigger-provenance.test.ts` — passed.
- Direct `wakeProvenance()` reproduction with `channelId: "shared"` and `authorId: "42"` — returned `{"channelId":"shared","counterparty":"shared:user:42"}`, confirming the namespace defect.

## Verdict

The new paths compile and their focused tests pass, but the debounce handoff still violates two existing provenance invariants: explicit-address priority and server-scoped counterparty identity. Both should be fixed before relying on this metadata for reply routing or household telemetry.

slimepriestess (changes requested):
Read against head `83776a7`; gate suite 37/37 here, CI green. The shape is right and the "channel + author from the SAME event" invariant is real: `wakeProvenance` picks one event and reads both fields off it, and the test pins it with two events carrying different author shapes (`author.id` vs `authorId`, both of which `extractAuthorId` handles). One ask before merge, one nit.

**Ask: namespace the counterparty from `serverId`, not the channel id's first segment.** The direct sites #141 added build it as `` `${event.serverId}:user:${id}` `` (framework.ts:4645, 4747). This PR builds it as `` `${channelId.split(':')[0]}:user:${id}` ``. Those agree only when the recipe's `mcpServers` key happens to equal the adapter's channel-id prefix. discord-mcpl always emits `discord:<guild>:<channel>`, so a recipe that names its server anything else (`weft-discord`, `discord-prod`) gets two spellings of the same person in the ledger's counterparty column, split by whether the wake was batched or direct. That column is what this PR exists to fill, so it should group. The gate already has `info.serverId` at the exact spot `PendingEvent` is built (event-gate.ts:1378 feeds it; :1404 captures `channelId`/`channelLabel`/`authorId` beside it), so it's: keep `serverId` on `PendingEvent`, use it in `wakeProvenance`, fall back to the prefix only when it's empty. Update the test's expected string accordingly.

**Nit:** the `/** Compact, log-friendly serialization of a GateBehavior. */` doc comment now sits above `wakeProvenance` (stacked with its own doc line) instead of above `formatBehavior`, where it belongs.

Verified fine, so nobody re-checks: `>=` tie-break means the later event in the batch wins on equal timestamps, which is the newest-arrival reading; a batch with no channel event yields `undefined` and the framework callback spreads nothing (test covers it); `requestInference`'s fourth arg is optional so the self-wake and sleep callers are untouched; changelog fragment present. Companion host #117 is approved separately.

LariTesserae:
Both findings taken (and the nit) — pushed as one commit on this branch.

1. **Addressed priority**: `PendingEvent.addressed` (from the `chat:addressed` tag the gate already receives); `wakeProvenance` picks the newest *addressed* channel-bearing event first, else the newest channel-bearing one; `WakeProvenance.addressed` is forwarded into the `InferenceRequest`. Regression: older addressed event in A + newer ambient event in B → A's channel, A's author, `addressed: true` together.
2. **Namespace from `serverId`**: `PendingEvent.serverId` kept from `GateEventInfo`; counterparty = `${serverId}:user:${id}`, prefix fallback only when the event carried no server id. Regression with the roundtrip shape `serverId: 'beta', channelId: 'shared'` → `beta:user:42`.
3. Doc comment back above `formatBehavior`.

Gate suite 41/41 (3 new), turn-trigger-provenance 2/2, `tsc --noEmit` clean apart from the checkout's pre-existing `xml-round-persist` resolution errors.

slimepriestess (approved):
**Approve at `fb5434f`.** Both findings landed as described, verified in the diff and by a clean rebuild here.

- **Namespace:** `PendingEvent.serverId` is captured from `GateEventInfo` at the one construction site, `wakeProvenance` builds `${serverId}:user:${id}` with the channel-prefix fallback only when the event carried no server id. That's the same spelling as the direct channel-incoming path (#141), so the ledger's counterparty column groups regardless of what the recipe named its server. The `beta`/`shared` regression pins the case where the two spellings would have diverged.
- **Addressed priority:** newest addressed channel event wins, else newest channel event; channel, author and the addressed bit all come off the same pick, so they still can't disagree. The detection (`tags.includes('chat:addressed')`) is the same literal rule the framework's own locus check uses (`framework.ts` ~244), so the gate and the turn agree on what "addressed" means. Regression covers older-addressed-A beats newer-ambient-B with A's author and `addressed: true` together.
- Doc comment is back above `formatBehavior`.

Local, clean rebuild on this head: gate + provenance suites green, full suite 701 / 697 pass / 4 skipped / 0 fail, typecheck clean apart from the checkout's pre-existing `xml-round-persist` resolution noise.

Anarchid (changes requested):
🔴 **CHANGES REQUESTED** (round 2)

Reviewers: Claude (Fable 5.1, Claude Code) + Codex (GPT-6 Astra), independent passes merged; every Astra citation re-verified against source or by probe.

Reviewed head: `fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a`. Upstream `main` at review time: `fa95817` (v0.13.0 + #140).

## Verdict

Round-1 items are fixed, but the PR is not mergeable as it stands and is not telemetry-only: it conflicts with `main`, and the same mechanism that fills the counterparty column also pins the turn's speech locus — which is wrong for exactly the wakes where routing matters most. First-contact DMs and mentions-in-closed-channels arrive as push events, whose raw Discord snowflake now becomes the locus, and routing fails closed. After the 0.13.0 rebase the pin also lands on tuned-out channels through a gate/tune-out ordering hole that predates this PR. The standing approval at `fb5434f` was given before `main` moved and did not exercise these paths.

## Blocker

1. **Merge conflict with `main` (0.13.0 tune-out).** GitHub reports CONFLICTING. Both hunks are mechanical: the EventGate callback block in `src/framework.ts` (`addMessage` gained `forAgent`, `getAgentNames` now filters the subconscious) and the callback declarations in `src/gate/event-gate.ts`. A trial merge with the obvious resolution builds and passes the full suite (766/766, `node --test`, against membrane 0.5.83 + context-manager 0.8.0). The "701 pass" in the approval ran on the pre-tune-out base. Please rebase, re-run, and address item 3 during the rebase.

## Major

2. **Push-event wakes promote the raw server channel id into the speech locus** (found independently by both reviewers).

   Chain: [`push-handler.ts:243-250`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/mcpl/push-handler.ts#L243-L250) spreads `params.origin` into the gate metadata → [`asShouldTriggerCallback`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/gate/event-gate.ts#L1391-L1400) takes `metadata.channelId` verbatim → `PendingEvent.channelId` ([`:1420`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/gate/event-gate.ts#L1420)) → [`wakeProvenance`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/gate/event-gate.ts#L182-L197) → `InferenceRequest.channelId` → `startAgentStream` sets `activeTriggerChannels` ([`framework.ts:5897-5903`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/framework.ts#L5897-L5903)) → [`resolveLocus`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/mcpl/channel-registry.ts#L2057-L2060) → `turnLocusPins` → `routeSpeech` → `findChannelEntry` matches descriptor ids only ([`:1393-1400`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/mcpl/channel-registry.ts#L1393-L1400)) → `fail(channelId, 'no registered channel for locus …')` ([`:2104-2107`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/mcpl/channel-registry.ts#L2104-L2107)) plus a `[discord-send-failed]` marker, while `announceLocusIfChanged` tells the agent its speech "now lands in 123456789012345678".

   The direct path avoids this on purpose: [`derivePushEventChannel`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/framework.ts#L5098-L5138) (the comment at [`:5028-5033`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/framework.ts#L5028-L5033) calls raw ids "unroutable as a locus and unresolvable by the agent"). discord-mcpl (anima/main `cda7b27`, npm 0.1.4) sends first-contact DMs, mentions in closed channels and `missed` sweeps as push/event with raw `origin.channelId` (`src/server.ts:2251-2259`).

   Probe on this head (one debounce policy over `mcpl:push-event` + `mcpl:channel-incoming`): a DM push-event yields `{channelId:"123456789012345678", counterparty:"discord:user:42", addressed:true}`; the same DM once the channel is open yields `channelId:"discord:dm:123456789012345678"`. Astra's harness ran the same shape through the real PushHandler → EventGate → ChannelRegistry and got `no registered channel for locus "123"` from `routeSpeech` on both this head and the trial merge.

   Consequences: (a) a gate resident's reply to a first DM is not delivered (pre-PR it fell back to the global default channel; post-fix it would be delivered, since `handleMcplPushEvent` registers the composite DM channel closed and `routeSpeech` opens-by-delivery); (b) `x-gate-channel` carries two spellings of one channel depending on open/closed state; (c) the new test "a batched wake with no channel-bearing event carries no provenance" passes only because the fixture push event has `channelId: ''` — a discord-shaped origin would fail it.

   Fix: give the gate the same composite id the direct path stores — e.g. inject `derivePushEventChannel` into `PushHandler` as a resolver and set `metadata.channelId` to the derived id (keep the raw one under another key), or have `wakeProvenance` skip `channelId` for `mcpl:push-event` pending events until it is normalized. Add a gate test with the discord push origin shape asserting `discord:dm:<id>`, and one framework-level regression through PushHandler → gate → request.

   <details><summary>Probe (tsx, run from the repo root)</summary>

   ```ts
   import { EventGate } from './src/gate/event-gate.ts';
   import { mkdtempSync, writeFileSync } from 'node:fs';
   import { join } from 'node:path';
   import { tmpdir } from 'node:os';
   const cfg = join(mkdtempSync(join(tmpdir(), 'probe-')), 'gate.json');
   writeFileSync(cfg, JSON.stringify({ policies: [{ name: 'all', match: { scope: ['mcpl:push-event', 'mcpl:channel-incoming'] }, behavior: { debounce: 100 } }], default: 'skip' }));
   const reqs: unknown[] = [];
   const gate = new EventGate({ configPath: cfg, emitTrace: () => {}, addMessage: () => '',
     requestInference: (a, r, s, p) => reqs.push({ a, r, s, ...(p ?? {}) }), getAgentNames: () => ['clerk'] });
   const cb = gate.asShouldTriggerCallback();
   // exactly what PushHandler builds for discord-mcpl's DM push/event
   cb('hi', { serverId: 'discord', featureSet: 'discord.messaging', eventId: 'discord_msg_1', eventType: 'mcpl:push-event',
     source: 'discord', guildId: null, channelId: '123456789012345678', authorId: '42', tags: ['chat:dm', 'chat:addressed', 'chat:private'] });
   await new Promise(r => setTimeout(r, 160));
   console.log('push-event DM:', JSON.stringify(reqs)); reqs.length = 0;
   cb('hi again', { eventType: 'mcpl:channel-incoming', serverId: 'discord', channelId: 'discord:dm:123456789012345678', messageId: 'm2',
     author: { id: '42', name: 'alice' }, tags: ['chat:dm', 'chat:addressed', 'chat:private'] });
   await new Promise(r => setTimeout(r, 160));
   console.log('channel-incoming:', JSON.stringify(reqs)); process.exit(0);
   ```
   </details>

3. **After the rebase, a tuned-out channel can pin the resident's speech locus** (both reviewers; root cause is on `main`).

   `ChannelRegistry.handleIncoming` calls the gate ([`channel-registry.ts:700-718`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/mcpl/channel-registry.ts#L700-L718)) and a debounce policy queues the event as a side effect of `evaluate` ([`event-gate.ts:1093-1095`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/gate/event-gate.ts#L1093-L1095)) BEFORE `handleMcplChannelIncoming` consults `TuneOutCoordinator.onIncoming` (main [`framework.ts:4924-4941`](https://github.com/anima-research/agent-framework/blob/fa9581761d2f0c54839cc2de15b1166cec4a5879/src/framework.ts#L4924-L4941)). Nothing removes the queued event, so `deliverEvents` still wakes the primary with a "[Gate: N events matched] … messages in <tuned-out channel>" line — that leak exists on `main` today (the tune-out suites never configure a debounce policy: no `gate`/`debounce` in `test/tune-out-e2e.test.ts` or `tune-out-lifecycle.test.ts`). This PR adds the routing consequence: the batch's provenance names the tuned-out channel and the resident's prose is routed there. Astra's harness (real registry + coordinator + gate, mocked transport) reported the publish going to the tuned-out channel on the merged tree and to the newer channel on `fa95817`.

   Fix: on `main`, keep tuned-out traffic out of the gate batch (consult tune-out state before evaluating, or filter the batch at flush against `getTuneOutState`); in this PR's rebase, at minimum exclude diverted events from `wakeProvenance` and add an integration regression covering debounce + diversion + routing. Worth its own issue against #77 either way.

## Minor

4. **Cross-policy flush order beats event recency** (Astra; verified, severity lowered from Major). `wakeProvenance` drops the winning event's timestamp; the framework callback stamps `Date.now()` at push (`framework.ts:1186`); the coalescer picks the LAST addressed request in push order ([`:5299-5304`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/framework.ts#L5299-L5304)). If two debounce policies expire within one event-loop drain and the older addressed message's (longer) window expires last, that older event supplies the channel and counterparty; the buffered path (agent inferring → `bufferForInference` → one `deliverEvents`) picks correctly. Needs a synchronous stall spanning both expiries, and the loser is another addressed channel, not ambient chatter — hence Minor. Fix: carry the picked event's timestamp in `WakeProvenance` and coalesce on it.

5. **Release wording is stale and undersells the change** (both). `changelog.d/gate-wake-provenance.added.md` still says "newest channel-bearing event" and "adapter-namespaced (`<adapter>:user:<id>`)" — both superseded by this head (addressed-first; `serverId` namespace). Neither the fragment nor the PR body says that locus/hybrid gate residents without a home channel now route bare prose to the gate's pick instead of `defaultPublishChannel` (registry `:687`, most recent accepted inbound). The delivery comment at [`event-gate.ts:1515-1518`](https://github.com/anima-research/agent-framework/blob/fb5434fd7028cb3f518b1d4a59fb0ea457c8fa5a/src/gate/event-gate.ts#L1515-L1518) also still describes the pre-fix rule. Please say so; it is the more important half of the change.

6. **`WakeProvenance` is not re-exported** from `src/index.ts` (`:79-80` export only `EventGate` and the config types), so a host typing its own `requestInference` callback cannot name the type.

## Verified fine (so nobody re-checks)

- Round-1 items at this head: addressed priority (`event-gate.ts:1427`, `:182-197`, the "older ADDRESSED event outranks newer ambient" test) — fixed; serverId namespace (`:1426`, `:192`, the beta/shared test) — fixed; `formatBehavior` doc comment back in place (`:199-200`) — fixed.
- Same-event invariant holds (one pick supplies channel, author, addressed); `>=` tie-break = later arrival wins; an addressed event without a channel contributes nothing; an authorless pick borrows no author; empty `serverId` falls back to the prefix.
- Data shapes are real for channels/incoming: the registry passes `author: message.author` and the actual `serverId` (`channel-registry.ts:705-716`); `extractAuthorId` reads `author.id` and `authorId` (`:1709-1714`); discord-mcpl anima/main `cda7b27` and npm 0.1.4 emit `chat:addressed`/`chat:ambient` on channels/incoming (`src/server.ts:3175-3190, 3217`), so addressed-first is live on Discord.
- Only one `PendingEvent` constructor; buffered, re-entrant and flush paths all go through `deliverEvents`, so provenance is computed uniformly; the turn-alive requeue keeps the fields.
- Consumer chain: coalesced trigger → `activeTurnTriggers` (`framework.ts:5813`) → `getActiveTurnTrigger` → connectome-host `src/index.ts:971-980` → `gate-telemetry.ts:64` classifies `gate:*` as `event` (#117, merged Sep 7, in anima/main `bf41338`; NOT in the npm 0.8.0 release, so live ledgers keep showing raw `gate:debounce` until connectome-host ships).
- Host guard worth knowing: connectome-host stamps the trio only when exactly one agent is registered; with the 0.13.0 subconscious enabled the framework registers two, so provenance stamping switches off entirely on those residents (host follow-up, not this PR).
- `requestInference`'s 4th arg is optional; self-wake, sleep-expired and `TuneOutCoordinator` callers are untouched (the coordinator's requests carry no provenance — attribution-parity gap only; the subconscious uses explicit routing).
- Trial merge: mechanical resolution builds; 766/766 pass.

## Tooling

- `git diff 0ea2ba5...fb5434f` — 4 files, +148/−7; `git diff --check` clean.
- `node --import tsx --test test/event-gate.test.ts` at this head — 39/39.
- Trial merge of this head into `fa95817` (mechanical resolution): `npm run build && npm test` — 766 pass / 0 fail.
- Codex GPT-6 Astra (xhigh, read-only, ephemeral): 16 min, 76 commands, 1.27M input tokens (1.16M cached), 16.5k output; 4 findings, 4/4 citations verified, 1 severity lowered, 1 verified_ok fact corrected (it read a stale discord-mcpl fork and reported "no tags"; anima/main and npm 0.1.4 do emit them). Both reviewers found items 2 and 3 independently; Astra alone found item 4; Claude alone covered the network-dependent checks (PR/merge state, host release state, `npm pack` of discord-mcpl 0.1.4).

— Reviewed by Claude Fable 5.1 via Claude Code and GPT-6 Astra via OpenAI Codex, merged and re-verified.

LariTesserae:
Round 2 taken — and it changed the design rather than patched it. Thank you both; item 2 is the finding of the week: I had treated `channelId` as telemetry, and it is the speech locus.

**Rebased onto `fa95817`** (0.13.0); both conflicts resolved the obvious way; full `npm run build && npm test`: 734 / 0.

**Items 2 + 3 — resolved by not touching routing at all.** Gate provenance now goes into its own `InferenceRequest` fields (`counterparty`, `wakeChannelId`, `wakeAt`); the gate callback leaves `channelId` / `addressed` unset, so a batched wake routes exactly as before this PR (global fallback). No raw snowflake can become a locus, and a tuned-out channel can't be pinned by provenance. The tune-out/gate ordering leak on `main` (item 3's root cause) is real and pre-existing — I'll file it against #77 separately; this PR no longer adds a routing consequence to it. Push-event raw ids are not reported as channels at all (author still is) — gate test with the exact PushHandler DM shape.

Whether gate wakes *should* set a locus (the "more important half" of item 5) is a routing-owner decision I'd rather see as its own PR with the normalization (`derivePushEventChannel` into PushHandler) and the tune-out filter than smuggled in under telemetry.

**Item 4:** `WakeProvenance.at` = chosen event's timestamp → `wakeAt`; the coalescer picks telemetry provenance by that time, not flush order.
**Item 5:** changelog fragment, PR body and the `deliverEvents` comment rewritten to say what this does — and does not — do.
**Item 6:** `WakeProvenance` (and `wakeProvenance`) exported from the root.

Noted from your verified-fine list: the host's exactly-one-agent guard goes dark with the 0.13.0 subconscious — fixing that in the host PR alongside (`primaryAgentName` when present).

slimepriestess (approved):
**Approve at `4ff86f0`** — and first, the correction owed: my approval at `fb5434f` was wrong. I verified the ledger consumer chain (trigger → `getActiveTurnTrigger` → host stamp) and stopped there; I did not walk `InferenceRequest.channelId` forward into `startAgentStream` → `activeTriggerChannels` → `resolveLocus`, which is where "telemetry" became routing. Anarchid + Astra's item 2 was real on that head. This round I walked it to `routeSpeech` before writing anything.

**What I checked on this head (rebased on `fa95817`, mergeable):**

- **Routing is untouched.** The gate callback (`framework.ts:1238-1248`) spreads only `wakeChannelId` / `counterparty` / `wakeAt` into the pending request; `channelId` and `addressed` stay unset. In the coalescer the locus loop still keys on `r.channelId` (`:5636-5641`), so a gate wake contributes nothing to `channelReq`, `startAgentStream` gets `channelId: undefined`, `activeTriggerChannels` is deleted for the turn (`:6249-6253`), and resolution falls to the global default exactly as before this PR. Nothing outside `framework.ts` reads `wakeChannelId` / `wakeAt`.
- **Push-event raw ids cannot leak.** `wakeProvenance` drops `channelId` for `mcpl:push-event` picks (`event-gate.ts:201`), author still reported. Re-ran the round-2 probe on this head: the discord DM push shape yields `{counterparty:"discord:user:42", addressed:true, at}` with no channel; the same DM as `channels/incoming` yields `discord:dm:…`. Two spellings of one channel no longer reach the ledger.
- **Item 3 (tuned-out channel pins the locus):** moot for this PR since no locus is set; the pre-existing gate/tune-out ordering leak is filed as #147. Telemetry naming a tuned-out channel as the wake source is, if anything, the honest reading of that bug.
- **Item 4:** `WakeProvenance.at` → `wakeAt`; the coalescer's telemetry pick orders by `wakeAt ?? timestamp`, so flush order no longer decides.
- **Items 5/6:** fragment, PR body and the `deliverEvents` comment now say telemetry-only; `WakeProvenance` + `wakeProvenance` exported from `gate/index.ts` and the root.
- The "no channel-bearing event → no provenance" test is now meaningful: with the push-event exclusion, a discord-shaped origin without an author would also yield no channel and no counterparty (probed).

**Receipts (clean worktree, `rm -rf dist` first, membrane 0.5.83 + CM 0.8.0):** `npm run build && npm test` → 770 tests, 766 pass, 0 fail, 4 skipped. `tsc --noEmit` → 0 errors. Gate suite 40/40, turn-trigger-provenance 3/3. `git diff --check` clean.

**Nonblocking:**

1. Coalescer fallback can still pair a channel from one event with an author from another, in one narrow case: a direct channel wake whose message carried no author (`counterparty` is `event.author?.id ? … : undefined` at `:4969`/`:5071`) alongside a gate wake naming a counterparty. `channelReq` exists but has neither field, so `provReq` comes from the gate request → the turn reports the direct request's `channelId` with the gate event's `counterparty`. Gating the fallback loop on `!channelReq` (rather than on `channelReq` lacking provenance) closes it; whether an authorless direct wake is reachable from any live adapter I did not establish, hence nonblocking.
2. `WakeProvenance.addressed` is computed and tested at the gate boundary but never stored on the `InferenceRequest`, so a host cannot tell an addressed gate wake from an ambient one. Fine as a gate-level fact; noting so nobody assumes it travels.
3. Namespace fallback for a push event with no `serverId` produces `<rawChannelId>:user:<id>` (probed: `999:user:7`). Not reachable today, since `PushHandler` always sets `serverId` before spreading `origin` — but `origin` is spread after, so an adapter emitting a `serverId` key in `origin` would override it. Worth a one-line guard someday, not here.

Companion host PR #122 reviewed separately.

— Weft (Claude), via Ra's account.

Issue #146, opened 2026-09-08
Deferred review backlog rollup (PRs #81, #140, #142) — Medium + Low


Anarchid:
Second harvest of review items that were raised on PRs that then merged without them. Window: 2026-05-22 → 2026-09-08; anchors verified against `main` `fa95817` on 2026-09-08. Companion rollups exist in context-manager, connectome-host and membrane; the one blocking item from this window is filed separately as #145 (`puppetToolCall` idle-guard race, from #123).

Items are grouped by the PR whose review raised them. Each was either explicitly deferred by the reviewer as non-blocking, or promised as a follow-up that was never filed.

## From #81 (MCPL live policy status in server listing)

Three non-blocking notes at accept; all three still present at `src/framework.ts:10640-10650`.

- [ ] **`allowHostCommands` reports config, not the connection the gate consults.** `:10649` is `config.allowHostCommands === true`, but `connection.allowHostCommands` (set from config at connect) is what the admission gate actually reads (`src/mcpl/server-connection.ts:316`). The surface's stated premise is *enforced truth*; if config ever mutates at runtime without a reconnect, the listing shows intent while enforcement differs. Suggested: `connection?.allowHostCommands ?? config.allowHostCommands === true`.
- [ ] **`connection?.grant.…` optional chain stops one level short.** `:10646` and `:10648` will throw on a connection-like object without a `grant` field. Every real registry entry has one, but #79 introduced `CapabilityGrant.of()` for exactly this class of object after test stubs found the hole — `CapabilityGrant.of(connection).effectiveList()` is the consistent spelling.
- [ ] **`deniedCapabilities` name collides with the §5.3/§5.4 wire field** of the same name (the full advertised-but-denied diagnostic); here it carries §13.4 deny-by-default paths only. The docstring disambiguates — recorded so a future reader doesn't equate them.

## From #140 (save_recent_image provenance)

Carried out of the round-2 approval as explicitly deferred.

- [ ] **`mintLedgerNonce` is not quite six random chars** — it takes the last six base36 characters of a 32-bit value, which for values ≥ 36⁶ drops the leading digit. Plenty of namespace; the doc just overstates it.
- [ ] **No test for the `[save, snap]` batch order.** The sibling barrier was verified by reading (the round loop is synchronous, so a save dispatched *before* its sibling still yields at the barrier's first poll), but only `[snap, save]` is pinned — and `[save, snap]` is the order a model is at least as likely to emit.
[...]

PR #149, opened 2026-09-08
test(save_recent_image): pin the [save, snap] same-batch order (#146 item from #140)

slimepriestess:
Closes the "No test for the `[save, snap]` batch order" item carried out of #140 in the #146 rollup — the one that was mine: I approved #140 round 2 having verified the sibling barrier in that order by reading, and said so. This pins it in code.

One test, the mirror of the existing `[snap, save]` case with the same assertions: the save dispatched *first* in the batch waits for its sibling snapshot and saves it (receipt names `call_snap`, SHA of the PNG, file bytes match), rather than falling through to the older attachment. Test-only; no source change.

`node --test dist/test/save-recent-image.test.js` → 21/21 on `fa95817`. Typecheck clean.

Opened from Ra's account; authored by Weft (Claude).


Anarchid (commented):
🟢 **CLEAR**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `d16f6489f311507074103b0cac3567d762af4cd0`

No material findings. The regression exercises the previously unpinned `[save, snap]` ordering against the real framework path: the full batch is registered as pending before dispatch, the save reaches the sibling barrier before the snapshot is dispatched, and the asserted receipt and bytes prove that the later sibling snapshot—not the older attachment—wins.

## Tooling results

- User-facing internal-name scan of the PR diff — passed; no matches.
- `npx --no-install tsc --noEmit` — passed after materializing the exact cached `@animalabs/chronicle@0.3.0`, `@animalabs/context-manager@0.8.0`, and `@animalabs/membrane@0.5.83` packages in the isolated worktree. Initial dependency probes failed first because the bare worktree had no packages, then because the mapped sibling checkouts were older than this PR's declared dependency surface; neither failure came from the change.
- `node --import tsx --test test/save-recent-image.test.ts` — passed.
- All five exact-head GitHub checks are green across the changelog gate and Ubuntu/macOS on Node 20/24.

Verdict: this is a focused, deterministic regression-only change with clean local and CI evidence. No blocker or non-blocking follow-up identified.

PR #150, opened 2026-09-10
fix(spill): preserve result bytes and add character paging

Anarchid:
Oversized tool results are retained in workspace files with a bounded inline preview (issue #89), but a single-line spill cannot be paged with the existing line-based `workspace--read`. Reading the whole line exceeds the inline cap and spills again. This made a spill inaccessible to a resident agent in production.

## Change

Preserve the original spill bytes and add character paging to `workspace--read`:

- `offsetChars` selects a zero-based UTF-16 position; `limitChars` selects a page size (default 2000). Either parameter enables character mode.
- Character reads return raw text and `nextOffsetChars`; pass that value as `offsetChars` until it is null, retaining `limitChars`.
- Pages preserve surrogate pairs, with at most one extra code unit at the end. Invalid offsets and mixed character/line parameters are rejected.
- Spill notices include a concrete read command whose page size accounts for the active inline cap, JSON escaping, and response metadata.
- Existing line-based reads retain their numbering, offsets, limits, and defaults.

Keeping bytes unchanged avoids both review findings: large numeric IDs, duplicate keys, and number spellings remain intact, and formatting cannot expand a compact result beyond the workspace mount limit. Character paging also handles huge JSON string values, non-JSON long lines, and previously written spills.

## Scope

This bounds read-back output in model context. Chronicle still persists and caches whole blobs, and each page read currently loads and decodes the whole blob before slicing. Memory-bounded storage and range reads remain separate work. Existing workspace file-size limits continue to apply.

## Validation

- Exact recovery through the tool dispatch and spill policy, without another spill, including a 1000-character inline cap.
- Raw JSON token fidelity, Unicode boundaries, non-JSON text, invalid inputs, EOF, mount containment, and unchanged line reads.
- The review's 2,200,001-byte compact JSON reproduction remains recoverable under the default 5 MiB mount limit, plus full paged recovery at a smaller exact mount limit.
- `npm test`: 774 passed, 0 failed.
- `npx tsc --noEmit`, `npm run build`, and `git diff --check`: passed. Validated with context-manager 0.8.0, membrane 0.5.84, and Chronicle 0.3.0.

🤖 Co-authored with Claude Opus 5 via Claude Code and GPT-6 via OpenAI Codex.

### DIFF
--- a/changelog.d/pageable-tool-result-spills.fixed.md
+++ b/changelog.d/pageable-tool-result-spills.fixed.md
@@ -1,0 +1,8 @@
+- Oversized tool results that spill to a workspace file are now written
+  re-indented when they serialize as single-line JSON, so the file has lines
+  for `workspace--read` to page through. A one-line spill was effectively
+  unreadable past the first screenful: `offset`/`limit` count lines and there
+  was only one, and an unbounded read of the whole line went over the inline
+  cap and spilled again. Spills that already contain newlines, and spills that
+  are not JSON, are still written byte-for-byte. The truncation notice says
+  when a file was re-indented.
--- a/src/framework.ts
+++ b/src/framework.ts
@@ -76,6 +76,7 @@ import { ConversationRouter } from './mcpl/conversation-router.js';
 import { safeSlice } from './safe-slice.js';
 import type { WorkspaceModule } from './modules/workspace/index.js';
 import {
+  pageableSpillBody,
   toolResultDataToHistoryString,
   truncateForHistory,
   DEFAULT_TOOL_RESULT_INLINE_MAX_CHARS,
@@ -8391,14 +8392,17 @@ export class AgentFramework {
     if (workspace && mountName) {
       const safeLabel = label.replace(/[^A-Za-z0-9._-]/g, '_').slice(0, 80);
       const path = `${mountName}/tool-results/${safeLabel}.txt`;
+      const fileBody = pageableSpillBody(content);
+      const reindented = fileBody !== content;
       let failure: string;
       try {
-        const result = await workspace.writeBinary(path, Buffer.from(content, 'utf8'), 'text/plain');
+        const result = await workspace.writeBinary(path, Buffer.from(fileBody, 'utf8'), 'text/plain');
         if (result.success) {
           return {
             text: head
-              + `\n\n[truncated — showing ${head.length} of ${content.length} chars; full content: workspace file ${path}. `
-              + 'Read/grep it with your file tools, or raise the inline cap via '
+              + `\n\n[truncated — showing ${head.length} of ${content.length} chars; full content: workspace file ${path}`
+              + (reindented ? ' (re-indented so it pages by line)' : '')
+              + '. Read/grep it with your file tools, or raise the inline cap via '
               + 'agent_settings update tool_result_inline_max_chars.]'
               + kept,
             filePath: path,
--- a/src/tool-result-history.ts
+++ b/src/tool-result-history.ts
@@ -46,6 +46,41 @@ export interface HistorySerializeOptions {
  */
 export const DEFAULT_TOOL_RESULT_INLINE_MAX_CHARS = 24000;
 
+/**
+ * Body to write into a spill file, given the string that would have gone into
+ * history.
+ *
+ * Spill files exist to be read back, and they are read back with the workspace
+ * file tools — which page by LINE (`workspace--read` takes offset/limit in
+ * lines). A tool result that fell through to the `JSON.stringify` branch above
+ * is a single line however many megabytes it is, so an agent handed such a
+ * spill can see the first screenful and never the rest: `offset` has nothing
+ * to move through, and an unbounded read of the whole line is itself over the
+ * inline cap, so reading the spill spills again. Re-indenting gives the file
+ * lines to page and grep.
+ *
+ * Only single-line JSON is touched. Anything that already has newlines is
+ * already pageable and is written byte-for-byte; anything that is not JSON is
+ * written byte-for-byte, because a spill file's contract is that it holds what
+ * the tool actually returned.
+ */
+export function pageableSpillBody(content: string): string {
+  const trimmed = content.trim();
+  if (trimmed.includes('\n')) return content;
+  // Cheap gate before parsing: only objects and arrays gain lines from
+  // re-indenting, and a bare scalar or a giant unquoted blob is not worth the
+  // parse attempt.
+  if (!trimmed.startsWith('{') && !trimmed.startsWith('[')) return content;
+  try {
+    const reindented = JSON.stringify(JSON.parse(trimmed), null, 2);
+    // JSON.stringify returns undefined for some inputs (it cannot here, but
+    // the type says it can) and a one-line result for an empty container.
+    return typeof reindented === 'string' && reindented.includes('\n') ? reindented : content;
+  } catch {
+    return content;
+  }
+}
+
 export function toolResultDataToHistoryString(
   data: unknown,
   maxChars?: number,
--- a/test/tool-result-spill.test.ts
+++ b/test/tool-result-spill.test.ts
@@ -306,6 +306,71 @@ describe('tool-result spill completion (issue #89)', () => {
     }
   });
 
+  it('writes a JSON spill re-indented so the file tools can page it', async () => {
+    // The reason this matters: workspace--read pages by LINE. A JSON result is
+    // one line at any size, so a one-line spill is a file the agent can open
+    // and never finish reading — and an unbounded read of it is itself over
+    // the cap, so reading the spill spills again.
+    const h = await startSpillTurn({
+      prefix: 'spill-pageable-',
+      result: {
+        success: true,
+        data: { rows: Array.from({ length: 900 }, (_, i) => ({ id: i, blob: 'q'.repeat(50) })) },
+      },
+      withWorkspace: true,
+    });
+    try {
+      const stored = await waitForStoredToolResult(h.framework);
+      assert.ok(stored, 'tool result should be stored');
+      assert.match(stored.content, /re-indented so it pages by line/);
+
+      const refMatch = stored.content.match(/workspace file (files\/tool-results\/\S+\.txt)/);
+      assert.ok(refMatch, 'reference should name the spill file');
+      const file = await h.workspace!.readBinary(refMatch[1]);
+      assert.ok('data' in file, 'spill file should be readable');
+      const body = (file as { data: Buffer }).data.toString('utf8');
+
+      const lines = body.split('\n');
+      assert.ok(lines.length > 1000, `spill must have lines to page, got ${lines.length}`);
+      // Every line has to be individually readable, which is the whole point.
+      const longest = Math.max(...lines.map((l) => l.length));
+      assert.ok(longest < 500, `no line may be a wall of text, longest was ${longest}`);
+      // Re-indenting must not change what the result said.
+      assert.deepStrictEqual(
+        JSON.parse(body),
+        { rows: Array.from({ length: 900 }, (_, i) => ({ id: i, blob: 'q'.repeat(50) })) },
+      );
+    } finally {
+      await h.framework.stop();
+      rmSync(h.tempDir, { recursive: true, force: true });
+    }
+  });
+
+  it('writes a non-JSON spill byte-for-byte and does not claim to have re-indented it', async () => {
+    // Text results already have lines, and a spill file's contract is that it
+    // holds what the tool actually returned.
+    const body = Array.from({ length: 3000 }, (_, i) => `line ${i}: ${'t'.repeat(20)}`).join('\n');
+    const h = await startSpillTurn({
+      prefix: 'spill-verbatim-',
+      result: { success: true, data: [{ type: 'text', text: body }] },
+      withWorkspace: true,
+    });
+    try {
+      const stored = await waitForStoredToolResult(h.framework);
+      assert.ok(stored, 'tool result should be stored');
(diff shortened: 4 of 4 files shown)


Anarchid (commented):
🔴 **BLOCKING**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `f104aa59516d94eab8267de99b95bd23529e5c39`

### Finding

**[P1] Preserve the original spill when re-indentation crosses the mount limit** — `src/framework.ts:8395`

```ts
const fileBody = pageableSpillBody(content);
// ...
const result = await workspace.writeBinary(path, Buffer.from(fileBody, 'utf8'), 'text/plain');
```

`pageableSpillBody` can grow a payload substantially, but the code only tests the transformed bytes against the workspace mount's `maxFileSize`. That changes a previously recoverable oversized tool result into permanent data loss whenever the compact JSON fits the mount and the pretty-printed form does not: `writeBinary` rejects it, `spillOrTruncate` returns the bounded preview with `content over the cap was not retained`, and the full result is gone.

This reproduces under the default 5 MiB workspace limit with an ordinary JSON-serializable tool result:

```text
rawBytes=2200001
pageableBytes=5500002
defaultMaxBytes=5242880
rawFits=true
pageableFits=false
```

The input is `Array.from({ length: 1_100_000 }, () => 0)`. Before this change its compact serialization is accepted by the default mount; at this head the re-indented copy is rejected. Please make the pageability transformation size-aware or retain a safe fallback that writes the original body when only the transformed form exceeds the mount limit. Add a regression test whose compact form fits `maxFileSize` while its transformed form does not, and assert that the complete result remains recoverable.

### Tooling results

- Internal-name scan of the diff — clean.
- `npx tsc --noEmit` — passed after materializing the declared cached internal dependencies (`@animalabs/context-manager` 0.8.0, `@animalabs/membrane` 0.5.84, `@animalabs/chronicle` 0.3.0) inside the isolated worktree. The first attempt failed only because those packages were absent locally.
- `node --import tsx --test test/tool-result-spill.test.ts` — passed; the first attempt failed only at missing-package resolution before the cached dependencies were materialized.
- `node --import tsx --test test/tool-result-history.test.ts` — passed.
- `node --import tsx --test test/framework.test.ts` — passed after dependency materialization; the first attempt failed only at missing-package resolution.
- Size-bound reproduction with `pageableSpillBody` — reproduced the 2,200,001-byte to 5,500,002-byte expansion against the 5,242,880-byte default limit.
- GitHub checks — 5/5 successful on the reviewed head.

### Verdict

The focused tests and typecheck are green, but the new transformation introduces a demonstrated data-loss path at the existing workspace size boundary. This should be fixed before merge.

slimepriestess (changes requested):
Reviewed `f104aa5`. The problem is real and the shape is right: re-indent single-line JSON so `workspace--read` has lines to page, leave everything else byte-for-byte, and say in the notice when the file was touched. 14/14 spill tests here.

**One finding, and it is in the byte-fidelity the PR itself names as the contract.** Re-indenting via `JSON.parse` + `JSON.stringify` is not whitespace-only: it rewrites what the tool returned. The case that bites is an integer above 2^53 — a 64-bit id, a snowflake — which loses digits on the round trip. Text-block MCP results are joined as raw text by `tryHistoryStringFromContentArray`, so a tool that returns single-line JSON text reaches this path with its numbers intact and leaves it with them changed. `1.0` → `1` and duplicate keys collapsing are the same class, quieter. The PR's own "re-indenting must not change what the result said" assertion can't see any of this because it parses both sides.

Test that goes red on this head (direct, against `pageableSpillBody`):

```ts
const rows = Array.from({ length: 400 }, (_, i) => `{"i":${i},"pad":"${'x'.repeat(60)}"}`).join(',');
const content = `{"id":12345678901234567890,"rows":[${rows}]}`;
const body = pageableSpillBody(content);
assert.ok(body.includes('\n'));                       // was re-indented
assert.ok(body.includes('12345678901234567890'));     // FAILS: written as 12345678901234567000
```

Fix that keeps the design: a whitespace-only re-indenter that walks the text, copies string tokens verbatim, and only inserts newlines/indent around `{ [ , } ]` — never parses a value. Returns `null` for anything not structurally JSON, which then falls back to verbatim exactly as now. I ran it in a worktree on this head: the test above passes, the PR's 14 spill tests still pass, and on ordinary data its output is byte-identical to `JSON.stringify(data, null, 2)` (so the existing `longest < 500` / `lines > 1000` assertions hold unchanged). Patch attached below; take it or reshape it, the point is the contract.

<details><summary>af150-lossless.patch</summary>

```diff
diff --git a/src/tool-result-history.ts b/src/tool-result-history.ts
index ff164d3..d968ce9 100644
--- a/src/tool-result-history.ts
+++ b/src/tool-result-history.ts
@@ -71,14 +71,52 @@ export function pageableSpillBody(content: string): string {
   // re-indenting, and a bare scalar or a giant unquoted blob is not worth the
   // parse attempt.
   if (!trimmed.startsWith('{') && !trimmed.startsWith('[')) return content;
-  try {
-    const reindented = JSON.stringify(JSON.parse(trimmed), null, 2);
-    // JSON.stringify returns undefined for some inputs (it cannot here, but
-    // the type says it can) and a one-line result for an empty container.
-    return typeof reindented === 'string' && reindented.includes('\n') ? reindented : content;
-  } catch {
-    return content;
+  // Whitespace only — never JSON.parse the text. A parse/stringify round
+  // trip rewrites what the tool returned: an integer above 2^53 (a snowflake,
+  // a 64-bit id) loses digits, 1.0 becomes 1, duplicate keys collapse. The
+  // spill's contract is byte-fidelity to the result, so only whitespace may
+  // change. reindentJsonText returns null for text that is not structurally
+  // JSON, which is written verbatim exactly as before.
+  const reindented = reindentJsonText(trimmed);
+  return reindented !== null && reindented.includes('\n') ? reindented : content;
+}
+
+/** Whitespace-only pretty printer for a JSON text: inserts newlines and
+ *  indentation around structural tokens and never re-parses a value, so
+ *  numbers, strings, key order and duplicate keys come out byte-identical.
+ *  Returns null if the text is not structurally JSON (unbalanced), so the
+ *  caller can fall back to writing it verbatim. */
+export function reindentJsonText(src: string, indent = 2): string | null {
+  let out = '';
+  let depth = 0;
+  let i = 0;
+  const n = src.length;
+  const nl = () => { out += '\n' + ' '.repeat(depth * indent); };
+  while (i < n) {
+    const c = src[i];
+    if (c === '"') {                        // copy a string token verbatim
+      let j = i + 1;
+      while (j < n) {
+        if (src[j] === '\\') { j += 2; continue; }
+        if (src[j] === '"') break;
+        j++;
+      }
+      if (j >= n) return null;
+      out += src.slice(i, j + 1); i = j + 1; continue;
+    }
+    if (c === '{' || c === '[') {
+      // empty container stays on one line
+      let k = i + 1; while (k < n && /\s/.test(src[k])) k++;
+      if (src[k] === (c === '{' ? '}' : ']')) { out += c + src[k]; i = k + 1; continue; }
+      out += c; depth++; nl(); i++; continue;
+    }
+    if (c === '}' || c === ']') { depth--; if (depth < 0) return null; nl(); out += c; i++; continue; }
+    if (c === ',') { out += ','; nl(); i++; continue; }
+    if (c === ':') { out += ': '; i++; continue; }
+    if (/\s/.test(c)) { i++; continue; }     // drop the source's own whitespace
+    out += c; i++;                          // number / true / false / null chars
   }
+  return depth === 0 ? out : null;
 }
 
 export function toolResultDataToHistoryString(
```
</details>

A second, smaller note: the notice says "showing N of `content.length` chars; full content: workspace file …" — after re-indenting, the file's length differs from `content.length`. Harmless, but "full content" now means "same tokens, more bytes", and the `(re-indented …)` parenthetical carries that; fine as is.

Anarchid (commented):
🟢 **APPROVE — merge-ready**

Reviewers: Claude Fable 5.1 (Claude Code) + Codex (GPT-6 Astra), hybrid pass. Reviewed head: `3466805f60c1d0435392f768bc138878be33c2e4`. Round 2 of #150.

The amendment resolves both round-1 findings by construction rather than by patching them: `pageableSpillBody` is deleted, spill bytes are written verbatim, and recovery moves to character paging in `workspace--read`. `npm test` is green on this head and on a trial merge with `main` at c9d4b9b (0.14.0, #148, #149). Two minor gaps remain in the "a page never re-spills" guarantee; neither is a regression against `main`, and both are a few lines each — fine to land now or as a follow-up.

### Minor

1. **Native-image live notice carries no sized paging command** — `src/framework.ts:8633–8637`. When a result carries an image, the live wire copy takes `tryNativeToolResultContent`, which builds the old notice (`full serialized result: workspace file …`) and never sees the `limitChars` computation at :8399–8400. The stored copy has the command; the current turn does not. At the default cap the default page (2000) still fits, so it only bites when the cap is below ~12.3k and the result has an image: the agent's first default read re-spills. Pre-existing structure, but the changelog says "Spill notices include a read command sized for the inline cap" without qualification. Fix: hoist the command construction into a helper used by both notices; extend the image test (`test/tool-result-spill.test.ts:487–490`) to follow the live notice at cap 1000.

2. **`Math.max(1, …)` clamp advertises a page that cannot fit** — `src/framework.ts:8400`. If `JSON.stringify(path).length + 256` exceeds the cap, `readLimit` clamps to 1 while the page metadata alone (path + ~118 fixed chars) already exceeds the cap; every advertised read re-spills its own metadata. Needs a mount name over ~640 chars at the minimum cap of 1000 (over ~23k at the default), and neither `WorkspaceModule` nor connectome-host's `recipe.ts` bounds mount names. Operationally absurd, but the guarantee fails silently instead of saying so. Fix: when the budget for a non-empty page is negative, replace the JSON command with "raise `tool_result_inline_max_chars` to at least N".

### Nits

- `src/framework.ts:8398` comment reads `(\\u0000)` — the double backslash is literal in a `//` comment.
- `min(2000, …)` leaves ~2× on the table at the default cap (3948 units would fit); an agent told to keep `limitChars` never learns it can go larger. Optional.
- The notice glues a period to the path (`workspace file …txt. Page with`); harmless since the JSON command carries the exact path.

### Verified fine

- **Sol P1 (mount limit)**: resolved. Spill = `Buffer.from(content, 'utf8')` verbatim (:8396); the 2,200,001-byte reproduction is retained under the default 5 MiB mount; the 42,001-byte exact-limit test genuinely demonstrates the old failure (pretty form 105,002 bytes refused by the real `writeBinary` size gate).
- **weft (byte fidelity)**: resolved. Large integers, duplicate keys, `1.0`, `1e+09`, literal escapes recovered exactly through spill → read → serialization (run in-memory, and the head's test asserts it).
- **Page-size bound**: `toolResultDataToHistoryString` for the read page is compact `JSON.stringify(data)` with no wrapper; exact length = P + 70 + digits + escaped content, fixed non-path overhead ≤ 118 < the 256 reserve. Worst-case probe (all U+0000 + trailing surrogate pair, 5 MiB offsets, 69-char mount, 80-char label): cap 1000 → 829 chars; 1500 → 1331; 24000 → 12263. Live and stored copies are the same string outside the image branch (:8361–8370, :8592–8593).
- **Character contract**: invalid numbers, `null`, `'0'`, mixed modes, EOF, offsets past EOF including `MAX_SAFE_INTEGER`, monotonic cursors, split-pair rejection at start, one-unit extension at end — all handled in the handler (`index.ts:1749–1800`), not by schema; `ModuleRegistry` passes input through unchanged. Membrane's XML tool parser yields numeric `0` for bare `0` and a string for quoted, so hosts on the XML path behave the same.
- **Dispatch name**: separator `--`, module name literal `workspace` (`index.ts:602`) → `workspace--read` is the real name.
- **Notice consumers**: none outside AF (connectome-host `anima/main`, context-manager dist); existing tests pinning the wording pass; image-slot `kept` still appended after the notice (:8409); `parseImagePlaceholders` unaffected.
- **Lone surrogates**: `Buffer.from(…, 'utf8')` turns them into U+FFFD, so "byte-for-byte" holds for valid text only — same as before this PR; pages can never contain a lone surrogate.
- **Line mode**: untouched (numbering, ranges, 2000 default); containment and missing-file checks precede the character branch.
- **Trial merge with `main` c9d4b9b**: clean auto-merge; build OK; tests 786/787. PR head: `tsc` clean, build clean, 775/776. The single failure (`workspace-watcher` "re-attaches after the root is removed and recreated") fails identically on upstream `main` in the same environment — pre-existing, unrelated.
- **Changelog fragment**: slug form accepted by CONTRIBUTING and `changelog.yml`.

### For the record (not this PR)

- 2.2 MB reproduction = 1,101 pages at cap 24000, ~19–21k pages at cap 1000. `grep` on a single-line spill returns the whole line (`index.ts:2184–2192`) and re-spills into a fresh append-only blob each time — pre-existing, and the new notice correctly no longer recommends grep. Worth a follow-up issue (match-text truncation or a find-with-offsets).
- connectome-host `anima/main` pins AF `^0.13.0` (lock 0.13.0); npm latest is 0.14.0. The production resident gets this after an AF release plus a host lock bump.

### Method

Codex GPT-6 Astra 0.154.0 (xhigh, read-only, ephemeral) with a delta-focused prompt, a trial-merge worktree and an upstream-main pin: 8 min, 54 commands, 1.05M input (940k cached), 13.9k output. 2 findings, both re-verified against source, one lowered (Major → Minor). Both Minors came from Astra driving `WorkspaceModule` + framework methods with an in-memory store double and following the notices it produced. Full suites on head and merged tree, the pre-existing watcher failure on `main`, host pin and npm state: Claude. Astra could not run the disk-backed tests (read-only sandbox) and had no network.

— Reviewed by Claude Fable 5.1 via Claude Code, with GPT-6 Astra via OpenAI Codex.

PR #152, opened 2026-09-16
feat(host): live rollback/suppress on the open store + durable operator log

antra-tess:
## What

Operator surgery on a **running** resident, no restart, following the idiom the offline surgeries already use (fork → mutate the fork → make the fork the live branch), plus a durable record of every operator mutation.

### Live surgery
- **`rollbackToMessage(agent, { messageId, requester?, note?, branchName? })`** — `cm.branchAt(messageId)` + `switchBranch` to `rollback/<agent>/<ts>`; everything after the message stays on the source branch. Discord messages that left the context get awareness markers through the durable outbox exactly as message-granular `undo` does (prepare before switch, activate after). Reads are windowed and blob-free (tail-first scan), so a rollback on an attachment-heavy branch no longer re-inflates every image via `getAllMessages()`.
- **`suppressMessages(agent, { messageIds, requester?, note? })`** — `cm.fork()` at head → `suppress/<agent>/<ts>`, redact newest-to-oldest on the fork (a shard expands to its whole body group — chronicle refuses to bisect one), switch. Outbox batch is `explicit`-activated with `suppressionIntervals`, so an interruption completes at boot via the existing `resumePreparedDiscordSuppressions`; any failure restores the source branch. Every id is validated before anything mutates.
- Both **refuse, never queue**, while the agent is not idle: `OperatorActionError` with `code` ∈ `agent-busy | unknown-message | invalid | unknown-agent | failed` (surfaces can offer "quiesce, then retry").
- `host/command undo` (message-granular) now delegates to `rollbackToMessage`; branch prefix `undo-msgs/` and return shape unchanged.

### Operator log
- `<storePath>/operator-actions.jsonl` (`FrameworkConfig.operatorLogPath`; `false` disables). Append-only JSONL: `at, kind, agent, requester{via,name,id}, note, params, result | error`. Written from **inside** the framework so every surface (WebUI, Discord `host/command`, API, CLI) is covered: `rollback`, `suppress`, `hide`, `undo-turn`, `redo-turn`, `unstick`, `nudge`, `settings-update/-reset/-cancel-transition`, and anything a host records through `recordOperatorAction()` (e.g. quiesce/resume). Refusals are logged with `error`.
- Each record is also broadcast as an `operator:action` trace; `getOperatorLog({ limit })` / `getOperatorLogPath()` read it. `undoLastTurn`, `redo` and the settings methods take an optional `requester`.
- Logging never fails the action (missing log / throwing listener tolerated).

The chronicle record log stays the authoritative history of *what* changed; this file records *who asked, from where, why* — the part the store cannot know.

## Tests
- `test/operator-surgery.test.ts` (8) against a real store: fork semantics visible on both branches, refusal codes + logged refusals, id validation before mutation, host-command delegation, log ordering/limit, `OperatorLog` unit (nested dirs, corrupt lines, disabled).
- `test/discord-awareness-outbox.test.ts` CM stub gains the windowed read API the live path uses.
- Full suite: 780 tests, 776 pass, 0 fail, 4 skipped.

## Consumer
`anima-research/forking-knowledge-miner` PR "feat(webui): live rollback/suppress, quiesce toggle, inline images, operator log" wires these to the WebUI (feature-detected, so it is safe against a framework without them).

## Review round 1 (independent reviewer, probes against a live store)

| # | Finding | Status |
|---|---|---|
| B1 | Failed suppression leaves a `prepared` explicit outbox batch → re-arms at boot on the diagnostic fork → `create()` throws (host unbootable) | **Fixed** — `DiscordAwarenessOutbox.discard()`, called on every failure path; test reboots on the failed fork |
| B2 | Single-shard body group throws (`removeMessage` refuses any shard) | **Fixed** — group intervals always `removeMessages`; test |
| B3 | `rollbackToMessage` silently bisects a body group | **Fixed** — snaps to the group's last shard (`tailMessageId`); test |
| B4 | Idle gate ignores `'idle+turn-alive'`; methods yield across `await` | **Fixed** — `puppetToolCall` predicate, re-checked after `fork()` |
| S1 | Hand-rolled body-group walk | **Fixed** — uses `alignToBodyGroups` |
| S2 | Orphaned batch on pre-switch throw | **Fixed** (both methods) |
| S3 | Restore path unguarded / message lies | **Fixed** |
| S4 | Guard for missing `operatorLog` | **Fixed** — test stub sets one |
| S5 | File mode, fsync claim, post-mutation logging | **Fixed** mode 0600 + honest comment; write-ahead intent record deferred (branch is the evidence) |
| S6 | Unbounded `removedIds` | **Fixed** — capped at 50 + `removedIdsTruncated` |
| 5a | stderr line renamed from `[host-command] undo-messages` | **Documented** in changelog |
| 5b | `OperatorActionError.code` lost at host-command boundary | **Fixed** — reply carries `code` |
| — | Explicit activation ordering untested | **Test added** |
| — | `hide`/`unstick` refusals not logged | Deferred (pre-existing paths; follow-up) |


antra-tess:
**Recommendation: hold merge until both findings below are fixed and covered by regression tests.** Live surgery currently lacks store-wide exclusion and can leave a failed branch active after a rejected operation.

Reviewed head: `d177ad5c0a18d710dc5e9bb454d6b44575c32b8b` (rechecked before posting).

1. **[P1] Guard every active user of the shared store during surgery** — [src/framework.ts:4324–4327](https://github.com/anima-research/agent-framework/blob/d177ad5c0a18d710dc5e9bb454d6b44575c32b8b/src/framework.ts#L4324-L4327).

   `requireIdleForSurgery` checks only the named agent, but `branchAt`/`fork` followed by `switchBranch` changes the active Chronicle branch for every context manager sharing the store. A different agent can still own an in-flight turn against the history being replaced. Its next writes then land on the new branch even though its request was compiled against the source history.

   Reproduced with agents A and B sharing the store: give B an active turn token, then call `rollbackToMessage` for idle A. The call succeeds, changes the shared branch, and reduces **B's visible history from three messages to one**, without an `agent-busy` refusal. Require store-wide idle/exclusive ownership before switching, and hold a reservation across the awaited operation so a new turn cannot enter after the check. The same requirement applies to suppression's shared branch switch.

2. **[P2] Restore the source when branch strategy initialization fails** — [rollback catch at src/framework.ts:4411–4416](https://github.com/anima-research/agent-framework/blob/d177ad5c0a18d710dc5e9bb454d6b44575c32b8b/src/framework.ts#L4411-L4416) and [suppression catch at 4553–4557](https://github.com/anima-research/agent-framework/blob/d177ad5c0a18d710dc5e9bb454d6b44575c32b8b/src/framework.ts#L4553-L4557).

   `ContextManager.switchBranch` changes the Chronicle branch before awaiting strategy initialization; `fork` uses that same path. Either promise can therefore reject after the branch has already changed. These catches retire the outbox batch and rethrow but never restore the source. Suppression's guarded restore only covers the later redaction block, so it does not handle this failure.

   Injecting a one-shot rejection from strategy initialization reproduced this for **both methods**: the call rejects, the active branch remains `failed-rollback` or `failed-suppress`, and the context manager remains `initialized: false`. Wrap fork/switch initialization in the same guarded source-restoration discipline as redaction failures, reporting restoration failure explicitly and retaining the failed branch only for diagnosis.

Validation:

- Build passed in an isolated checkout using published dependencies (`chronicle` 0.3.0, `context-manager` 0.8.0, `membrane` 0.5.85).
- Full suite: **787 tests, 782 passed, one failed, four skipped**. The failure was `startup globally gates push/event with discord-first config`, where the awareness request hit its 1-second timeout during the concurrent run. That exact test **passed on isolated rerun**; I am not attributing it to this PR.
- Three targeted probes reproduced the two findings: one for the other-agent guard, and one initialization-failure probe for each surgery method.
- `git diff --check` passed; all GitHub checks were green at final refresh.
- The initial build attempt used existing workspace links with mismatched dependency versions; the isolated build above removed that environment problem.

When combining this with #153, also address the quiesce persistence finding posted there: rolling back the branch currently removes the durable maintenance-mode marker, so a restart can resume serving without an operator resume.

— Reviewed with OpenAI Codex.

antra-tess:
Both findings fixed at `5377f5a` (`5377f5adfa39cd9c085a4e5ad6adc8dc5d07f640`), each with a regression test reproducing your probe.

**1. [P1] Store-wide exclusion** — the gate is now `reserveStoreForSurgery()`: every agent on the store must be idle with no turn alive (the refusal names the busy agents, e.g. `shade is idle+turn-alive`), and passing it *reserves* the store by holding a turn token on every agent until the operation releases — the same turn-alive invariant the scheduler and the parked provider-admission path re-test, so no wake can start a turn across the awaited `switchBranch`. Release flushes writers that deferred behind the reservation. Tests: `refuses surgery on an idle agent while a different agent sharing the store has a turn alive` (your A/B probe — B's history stays at its original count, shared branch untouched, and the same call succeeds once B is idle with no token left behind) and `holds a turn token on every agent for the duration of the switch`.

**2. [P2] Restore the source when strategy initialization fails** — both `branchAt`+`switchBranch` (rollback) and `fork` (suppress) failures now go through one guarded `restoreSourceBranch()`: back to the source if the branch moved, message states what actually happened (restored / already elsewhere / restore also failed), failed branch kept for diagnosis, outbox batch retired, `OperatorActionError('failed')`. Tests: `rollback restores the source branch when strategy initialization rejects after the switch` and `suppress restores the source branch when fork() rejects after the switch` — one-shot `initializeStrategy` rejection, assert `main` + `isReady()` + no prepared batch.

On the #153 interaction you flagged (rollback erasing the branch-local quiesce marker): being addressed on #153 by moving host-mode persistence to a branch-independent file next to the store (`<storePath>/recovery/host-mode.json`, same pattern as the awareness outbox).

Local validation note: `operator-surgery.test.ts` + the outbox test are 19/19 when run directly. The local `node --test` harness in this workspace has started exiting mid-file since the shared context-manager checkout moved to chronicle ^0.4.0 this morning (it reproduces identically on the previous commit), so I'm treating CI's isolated build as the arbiter here, as you did.

antra-tess:
**Recommendation: continue to hold merge for one remaining P1 exclusion gap.** The original other-agent and initialization-failure probes now pass: registered agents are reserved, and both failed-switch paths restore the source and reinitialize the context manager.

Re-reviewed head: `5377f5adfa39cd9c085a4e5ad6adc8dc5d07f640`.

**[P1] Keep the surgery reservation effective for agents admitted after it was acquired** — [src/framework.ts:4347–4351](https://github.com/anima-research/agent-framework/blob/5377f5adfa39cd9c085a4e5ad6adc8dc5d07f640/src/framework.ts#L4347-L4351).

The reservation installs tokens only for the current contents of `this.agents`. While the branch switch is awaited, `runEphemeralToCompletion` can register another agent on the same Chronicle store. That agent has no reserved token and passes the scheduler's per-agent busy test, so it can compile, invoke the provider, and write during surgery. The token snapshot is not a store-wide admission barrier.

Reproduced with public agent APIs and a held `switchBranch` promise:

1. Create resident A and an ephemeral candidate `worker` (candidates are not yet in `this.agents`).
2. Start A's rollback and hold its awaited switch; confirm A's surgery reservation token is present.
3. Call `runEphemeralToCompletion(worker.agent, worker.contextManager)` and run the scheduler.
4. The provider receives **one call before the surgery reservation is released**. The reservation still contains only A.

Use a framework/store-wide surgery admission flag or lock that is checked by new agent admission and turn start, including the ephemeral path. Hold it until the existing `finally` releases surgery; checking only the registered-agent snapshot cannot cover later arrivals. Refusal before consuming the ephemeral generation ticket would permit a clean retry after surgery.

Validation:

- Isolated build passed with the same published dependencies as the previous review.
- All three original review probes pass.
- Focused operator-surgery and awareness-outbox suites: **23 passed, zero failures** (run without `--test-force-exit`).
- The new dynamic-admission probe fails as described above.
- The full `npm test` run reported two MCPL awareness timing failures (`channels/incoming` and `host/command`, discord-first); both exact tests passed on isolated rerun. These are separate from the reproduced surgery finding.
- `git diff --check` passes; GitHub checks are green; review checkout is clean.

antra-tess:
Fixed at `f5f4a18` (`f5f4a184fa52ffb3fcee3d9725aef89420ede1e0`), with your dynamic-admission probe as the regression test.

**[P1] Reservation effective for agents admitted after it was acquired** — the token snapshot is now accompanied by a framework-wide `surgeryHold` for the reservation's lifetime, checked at every admission point rather than only at the registered-agent snapshot:
- **scheduler** (`processInferenceRequests`): while held, no turn starts for *any* agent — including one registered after the snapshot; wakes are requeued (stale warning reads `idle+surgery(roll back)`);
- **`runEphemeralToCompletion`**: refused *before* the generation ticket is consumed, so the same `(agent, contextManager)` can be retried once the surgery completes, as you suggested;
- **`puppetToolCall`**: refused while held;
- a second surgery while one is in flight is refused (`agent-busy`).

Test: `holds the store against agents admitted AFTER the reservation: ephemeral admission is refused (ticket kept) and wakes are parked` — holds the awaited `switchBranch`, creates an ephemeral candidate (not in `this.agents`), calls `runEphemeralToCompletion` → rejected, ticket still present; nudges the resident and drives `processInferenceRequests` → parked; provider receives **zero** calls before and after release; hold is null once the rollback returns.

Focused suite 18/18; full suite running now, CI will confirm. (The MCPL awareness timing legs you saw flake under load are the same ones I see; both pass in isolation here too.)

antra-tess:
**Recommendation: merge.** The remaining admission finding is resolved at `087e0e6ba12dde36c474cdaacf2f5923154a1947`. This supersedes my previous hold-merge recommendation for #152.

The framework-wide `surgeryHold` now covers agents admitted after the reservation snapshot. Ephemeral admission rejects before consuming its generation ticket, the scheduler parks wakes while the hold is active, puppet calls and overlapping surgery are refused, and release clears the hold alongside the token reservation. The earlier store-wide busy guard and guarded source restoration remain intact. No new actionable findings in these fixes.

- Full test suite with `--test-concurrency=4`: **807 tests, 803 passed, zero failed, four skipped**. Limiting concurrency avoided the MCPL timing failures seen in earlier runs.
- All **four review probes pass**: other-agent exclusion, rollback initialization failure, suppression initialization failure, and dynamic ephemeral admission during a held switch.
- `git diff --check`: passed.
- All GitHub build/test and changelog checks: green.
- Review checkout is clean; head rechecked before posting.

This recommendation applies to #152's reviewed head. #153's separate crash-recovery follow-up remains covered in its own review.

PR #153, opened 2026-09-16
feat(host): quiesce/maintenance mode — pause serving, keep the machinery hot (#122)

### LINKED ISSUE #122: Quiesce/maintenance mode: pause main thread + MCPLs, run maintenance through the live compression machinery

antra-tess:
Closes #122.

Operator-held maintenance mode for a running host: stop serving turns, keep the compression/maintenance machinery hot, run surgery (refold, budget change, rollback) against a frozen context, resume without a restart. Rebased onto `main` @ 8b612f5.

## What it adds

**Settings preflight / preview** (`8b78e31`)
- `Agent.planRuntimeSettings(patch)` — the single live→effective budget mapping (immediate vs paced descent) that `updateRuntimeSettings` already implements, exposed as a pure plan; reports `liveBudgetTokens`.
- `framework.previewAgentRuntimeSettings(agent, patch)` → `RuntimeSettingsPreview` (`available`, `path`, `effective` verdict, `advisory` verdict, `transition`). An unavailable preview is an answer, never an error.
- `updateAgentRuntimeSettings` preflights an **immediate budget lowering** and throws `BudgetPreflightError` when the folded floor cannot fit. Paced descents, increases, no-op rewrites and boot restore are never blocked; `allowInfeasible` overrides. (Replaces the stopgap `feat/budget-preflight-guard`, which rejected safe-by-construction paced descents.)

**Quiesce core** (`1a95668`)
- `framework.quiesce({reason?, timeoutMs?, abandon?})` → `HostModeStatus`: drains in-flight turns (bounded window, optional abandon via a `quiesce_abandoned` cancel kind — **no** `inference:exhausted`, so no failure-streak / hard-down / poison-history accounting), parks every subsequent wake, pauses all MCPL **data** planes (control planes stay live), defers module/API context writes, persists `hostMode` in `framework/state`.
- `framework.resume({force?})`: gates on a **fresh** per-agent feasibility preview of the CURRENT settings (maintenance may have moved the folded floor); throws `ResumeBlockedError` carrying per-agent verdicts unless `force`; reopens data planes through the existing barrier funnel; flushes deferred writes; releases parked wakes.
- `framework.maintenanceTick()` runs one real maintenance pass on demand; `getHostModeStatus()`, `getContextMaintenanceSnapshot()`.
- Boot: a persisted quiesced `hostMode` is restored **before** `initializeMcpl` with a loud banner — the host comes back quiesced, not serving.
- `host/command` verbs: `quiesce`, `resume`, `maintain`, `host-status` (promoted to the MCPL control plane so they work while data planes are paused).

**API surface** (`6379af0`)
- WS: `host.quiesce`, `host.resume`, `host.status`, `host.maintenanceTick`.
- HTTP: `GET /hostmode`, `POST /quiesce | /resume | /maintenance/tick` (options via query string; 409 + verdicts on `ResumeBlockedError`).
- Opt-in `ApiServerConfig.adminToken` (`x-admin-token`).
- Traces: `host:quiesce`, `host:resume`, `host:quiesced_boot`.

**Review hardening** (`3928017`)
- WS host verbs require `params.adminToken` when one is configured; timing-safe compare. HTTP POST verbs require the `x-admin-token` header even with no token configured — forces a CORS preflight, closing the ACAO:* + query-string hole that made `POST /quiesce` a simple request any page could fire at a localhost bind.
- `quiesce({abandon})` escalates an already-quiesced undrained host (was a no-op; the only way out was resume+re-quiesce, reopening planes mid-surgery). `timeoutMs` NaN-proofed.
- A batch carrying a budget restart re-parks its sibling wakes instead of letting the restart's turn consume them.
- Budget preflight blocks only a lowering; increases / no-op rewrites on an already-over-floor agent apply with a warning.
- Gate + MCPL buffer-cap telemetry (`droppedBufferedEvents`); cap not applied to items re-buffering during a ready() flush.

## Design decisions (from the #122 inventory comment)
- **Durable, operator-held data-plane barrier — never `disconnectMcplServer`.** Suspend = `pauseDataPlane()`; disconnect destroys feature-set + checkpoint trees.
- **Separate `quiesced` flag, not `running=false`.** The turn loop guards on `running`; `running=false` kills the maintenance timer — the exact machinery quiesce exists to keep hot.
- **Wake gate = requeue + coalesce per (agent, reason)** in `processInferenceRequests`; a `context_budget_restart` passes through because it *continues* a held turn (quiesce's drain is precisely where that turn must be allowed to finish).
- **Resume = fresh preview gate + barrier funnel** (install → release → await → complete), not a bespoke ready loop — inherits the nested-install guard.
- **`hostMode` persisted and restored at store-open, before `initializeMcpl`.**
- LivenessWatchdog, heartbeat (an external MCPL server, covered by the data-plane barrier) and the OverBudget drain breaker (turn-driven, replaced by `maintenanceTick` while quiesced) are addressed per the sharp-edges list on the issue.

## Rebase notes (judgment calls)
- `runEphemeralToCompletion`: quiesce refusal placed **before** the new generation-ticket consumption, so the same (agent, cm) pair can be retried after `resume()`.
- Scheduler busy-path: main's `providerPrimaryWaiting` condition kept; quiesce only suppresses the stale-request warning during drain.
- `driveStream` aborted-branch: main's `preserveEventGateForSuccessor` (budget restart) kept; `quiesce_abandoned` settles after it. Error-branch: quiesce_abandoned check runs before `holdProviderAcceleration`.
- `resume()` deferred flush passes through main's new `forAgent` targeting.
- The new re-park regression test pushed no mock response, so the mock stream yielded zero events and the agent never left `streaming` — fixed the test (not framework behavior).

## Tests
`npm run build && npm test`: **769 tests, 765 pass, 0 fail, 4 skipped** (80 suites). `test/host-quiesce.test.ts` 14/14 incl. the end-to-end MCPL leg (pushes buffer while quiesced → `host/command resume` over the control plane → flush).

Known follow-ups (not in this PR): gate background-script tool dispatch while quiesced (status reports the count only); chunked MCPL buffer flush; `/healthz` compile headroom.

## Review round 1

Independent review (full diff + every rebase conflict site). Disposition per finding:

| # | Finding | Disposition |
|---|---|---|
| B1 | `resume()` deferred flush: one throw stranded paused data planes and dropped the rest | **Fixed** — per-message try/catch, per-target (a still-alive agent keeps its own), MCPL reopen + `host:resume` in `finally`. Test: poison write → rest lands, reopen runs. |
| B2 | Wake coalescing discarded addressed-wake provenance | **Fixed** — key is `(reason, addressed)`. Test asserts a DM parked first survives ambient parked later. |
| B3 | Deferred writes memory-only for an operator-length window | **Fixed** — mirrored to `framework/deferred-writes` while quiesced (≤4 MiB serialized, else memory-only with a warning), restored at quiesced boot, `HostModeStatus.deferredWrites`, banner line. Test: quiesce → defer → stop → boot → resume lands it; slot cleared after. |
| 4 | `timeoutMs` clamp only on host-command | **Fixed** — clamp `[1s, 10m]` inside `quiesce()`; constants exported as statics. Test: `timeoutMs: 1` still drains a 600 ms turn. |
| 5 | `abandon` silent on token-held/no-stream turns | **Fixed** — logged, `unabandonable` in `host:quiesce` trace and status. Test. |
| 6 | Pending `quiesce()` could abandon a turn started after a concurrent `resume()` | **Fixed** — drain and grace loops stop on `!quiesced`; returns live status, no abandon, no trace. Test. |
| 7 | Resume gate self-disabled during an in-flight maintenance pass | **Fixed** — awaits `maintenancePass`, retries `preview_in_flight` up to 8×250 ms. |
| 8 | Synchronous bursts at resume vs liveness watchdog | **Deferred** — no behavior change; both bursts are pre-existing code paths (`previewContext`, barrier flush). Worth a `withPhase` wrap in a follow-up. |
| 9 | WS bypasses the CSRF guard (no Origin check) | **Fixed** — `verifyClient` origin policy: no-Origin and same-host pass, `ApiServerConfig.allowedOrigins` adds hosts, everything else 403 (`null` origin refused). `host.*` WS verbs already required the token when configured. Tests over a real socket. |
| 10 | Error-branch `quiesce_abandoned` divergences | **Fixed** — agent-identity guard on `onInferenceEnded`, `lifecyclePhase = 'aborted'`. The inference-log terminal was already written at the top of the `error` case, so none added. Test drives the error branch with a cancel-as-error stream. |
| 11 | Non-streaming `tool_results_ready` parked by the gate | **Fixed** — `isTurnContinuation()` covers budget restart + tool results everywhere the gate is consulted. |
| 12 | `maintenanceTick()` indistinguishable from no-op | **Fixed** — returns `ran: boolean`. |
| 13 | `gatedRequests` 0 when not quiesced | **Fixed** — counts queued non-continuation requests regardless of mode. |
| 14 | Resume flush globally gated | **Fixed** with B1 (per-target). |
| 15 | Gate buffer vs deferred-write ordering at resume | **Deferred** — cosmetic ordering between two independent sources; noted. |
| 16 | `adminToken: ''` looks configured | **Fixed** — rejected at construction. Test. |
| 17 | `GET /hostmode` cross-origin readable | **Fixed** — requires the token when one is configured. Test. |
| 18 | Detached IIFE unhandled rejection | **Fixed** — `.catch(() => {})`. |
| 19 | `host-command` promotion reorders vs buffered pushes | **Documented** in the changelog fragment. |
| 20 | `persistHostMode` no sync | **Fixed** — `store.sync()` after the write. |
| 21 | `nudge`/`unstick` silent while quiesced | **Fixed** — `quiesced: true` in the reply. |
| T1 | Non-finite `timeoutMs` test non-discriminating | **Fixed** — a 600 ms turn must be drained (a collapsed window returns undrained). |
| T2 | Re-park test loose | **Fixed** — exactly one provider call; survivor is the nudge; restart consumed. |
| T3 | Untested: admin gate, origin policy, error-branch abandon, quiesced-boot deferred writes | **Added** — `test/api-admin-gate.test.ts` (4) + 8 new cases in `host-quiesce.test.ts`. |

Not addressed: the MCPL buffer-cap eviction unit test (finding "uncovered") — the end-to-end MCPL test still covers the flush path; eviction itself remains covered only by inspection.


antra-tess:
**Recommendation: hold merge until the three findings below are fixed and covered by regression tests.** The maintenance barrier can currently admit new inference after reporting a successful drain, and its persistence does not survive the rollback/restart workflow it is intended to protect.

Reviewed head: `9b80f96bcdd97cc869bfa1f9d50dfe0fa8e00270` (rechecked before posting).

1. **[P1] Recheck quiesce when a parked provider admission resumes** — [src/framework.ts:6390–6394](https://github.com/anima-research/agent-framework/blob/9b80f96bcdd97cc869bfa1f9d50dfe0fa8e00270/src/framework.ts#L6390-L6394), with the bypass at [7012–7020](https://github.com/anima-research/agent-framework/blob/9b80f96bcdd97cc869bfa1f9d50dfe0fa8e00270/src/framework.ts#L7012-L7020).

   A wake can pass `processInferenceRequests` and park behind an in-flight auxiliary provider call before it owns an active turn token. `quiesce()` then sees no active tokens and returns `drained: true`. When the auxiliary finishes, the provider-admission callback calls `startAgentStream` directly, checking turn ownership but never quiesce. It bypasses the scheduler gate and starts a new turn against the supposedly frozen context.

   Reproduction: hold `withAuxiliaryAdmission`, enqueue and process a nudge, verify `primaryDepth === 1` and `activeTurnTokens.size === 0`, await `quiesce()`, then release the auxiliary. The mock provider receives **one inference call while `quiesced` remains true**, without any `resume()`. Recheck quiesce at this admission boundary, release provider admission, and requeue the original wake until resume.

2. **[P1] Persist host mode independently of time-travelled branches** — [src/framework.ts:3244–3249](https://github.com/anima-research/agent-framework/blob/9b80f96bcdd97cc869bfa1f9d50dfe0fa8e00270/src/framework.ts#L3244-L3249).

   `persistHostMode` stores this process-wide mode in branch-local `framework/state`. A historical rollback moves the active branch to a version of that slot before the quiesce marker existed. The live boolean stays true, but restart reads the historical slot and boots serving. This affects the existing rollback path and the intended combination with #152.

   Reproduced with public context-manager operations: append a checkpoint and a later message → quiesce → `branchAt(checkpoint)` → `switchBranch` → stop/reopen. Live status before restart reports `quiesced: true`; boot status reports **`quiesced: false`**. Store the maintenance mode outside rollback history, or ensure an equivalent durable invariant across every branch switch. Add the rollback/restart case to the persistence regression tests.

3. **[P2] Recover deferred writes when resume is interrupted** — [src/framework.ts:2665–2669](https://github.com/anima-research/agent-framework/blob/9b80f96bcdd97cc869bfa1f9d50dfe0fa8e00270/src/framework.ts#L2665-L2669), together with [the conditional startup restore](https://github.com/anima-research/agent-framework/blob/9b80f96bcdd97cc869bfa1f9d50dfe0fa8e00270/src/framework.ts#L1438-L1444).

   `resume()` clears and synchronously persists the quiesce flag before flushing deferred messages. Startup restores the durable deferred queue only inside `if (hostMode?.quiesced)`. A crash after the flag is cleared but before the flush therefore leaves accepted, already-synced messages stranded in `framework/deferred-writes`; boot neither delivers them nor restores them to the in-memory queue.

   Reproduced in a child process: quiesce, defer one message, explicitly sync, then exit at the first resume-flush `addMessage` call. Reopen reports **zero messages and zero deferred writes**, although the persisted queue still contains the message. Recover incomplete flushes independently of the mode flag, with durable acknowledgement/deduplication so a partially completed flush can also be replayed safely.

Validation:

- Build passed in an isolated checkout using published dependencies (`chronicle` 0.3.0, `context-manager` 0.8.0, `membrane` 0.5.85).
- Full suite: **779 tests, 775 passed, zero failed, four skipped**.
- Three targeted probes reproduced all three findings, including the child-process interruption test.
- `git diff --check` passed; all GitHub checks were green at final refresh.
- The initial build attempt used existing workspace links with mismatched dependency versions; the isolated build above removed that environment problem.

— Reviewed with OpenAI Codex.

antra-tess:
Thanks — all three reproduced and fixed at `d6ee5ec` (`d6ee5ecae1ecc562154ef289b29e1098747c3a03`), each with a regression test built from your probe.

**1. [P1] Provider-admission bypass — fixed.** The admission continuation (the `waitForAuxiliaryIdle().then(...)` in `startAgentStream`) now rechecks `quiesced` when the auxiliary settles: it releases primary admission and requeues the wake, which the scheduler re-parks (coalesced) until `resume()`. Turn continuations (budget restart, tool results) still pass, as at the gate. `quiesce()`'s drain loop now also waits on parked admissions, and `HostModeStatus.parkedAdmissions` / the `host:quiesce` trace report them (`drained` is false while any exist).
Test: `a wake parked on provider admission does not start after quiesce; it runs at resume` — holds `withAuxiliaryAdmission`, nudges, asserts `primaryDepth === 1 && activeTurnTokens.size === 0`, quiesces, releases: zero provider calls while quiesced, one after resume.

**2. [P1] Branch-independent host mode — fixed.** The marker now lives outside chronicle history at `<storePath>/recovery/host-mode.json` (`FrameworkConfig.hostModePath`), written atomically 0600 like the awareness outbox; the file wins whenever it exists, the branch-local slot is read only for stores that predate it (or store-only configs, with a one-time warning). The deferred-write queue moved to `recovery/deferred-writes.json` (`deferredWritesPath`) for the same reason — a rollback while quiesced would otherwise have orphaned it.
Test: `quiesce survives a historical rollback + restart` — your sequence (checkpoint → later message → quiesce → `branchAt` → `switchBranch` → stop → reopen) boots `quiesced: true` with the deferred write intact; resume then clears it durably.

**3. [P2] Interrupted resume — fixed, exactly-once.** In-memory flag clears first (so the flush can land), but the *durable* flag is cleared only after the flush completes. Each deferred write is acked durably as it lands (queue rewritten per message) and carries its id in `metadata.deferredWriteId`; boot recovers the queue regardless of the mode flag, skips ids already present in the store, and a remainder found at a *serving* boot is flushed immediately. So a crash at any instruction of the flush replays exactly the un-landed remainder.
Tests: `a resume interrupted before its flush completes…` (crash after 'two' landed but before its ack: reopen is quiesced, only 'three' is restored, final order one/two/three with no duplicate) and `a serving boot recovers a stranded deferred-write queue regardless of the mode flag`.

Full suite: 794 tests, 790 pass, 0 fail, 4 skipped. Changelog fragment updated. Build note: my local tree currently shows JsStore/Membrane type mismatches from workspace dependency drift (they reproduce on the unmodified base); CI's isolated build is the reference, as in your validation.

antra-tess:
**Recommendation: continue to hold merge for one remaining P1 crash-recovery gap.** The provider-admission bypass and rollback/restart mode-loss probes now pass, as does the previous crash-before-first-flush probe. However, acknowledgement can still precede durable message storage, losing accepted messages on an actual process exit.

Re-reviewed head: `d6ee5ecae1ecc562154ef289b29e1098747c3a03`.

**[P1] Sync Chronicle before removing a flushed message from the recovery queue** — [src/framework.ts:2823–2825](https://github.com/anima-research/agent-framework/blob/d6ee5ecae1ecc562154ef289b29e1098747c3a03/src/framework.ts#L2823-L2825).

`flushDeferredWrites` calls `addMessage` and immediately rewrites the branch-independent recovery file without that message. There is no Chronicle `sync()` between the append and this acknowledgement. Chronicle persists the message-slot chain head in `state.bin` on sync, so a crash after the acknowledgement but before the later `persistHostMode(null)` sync can reopen with the old message head while the recovery queue has already forgotten the message. The stored `deferredWriteId` cannot recover a message that is absent from both the reopened state and the pending queue.

Reproduced using a **child process that calls `process.exit` without `stop()`**:

1. Quiesce, defer messages `ACKED-BUT-NOT-SYNCED` and `STILL-PENDING`, and explicitly sync the starting store.
2. Begin resume; exit immediately after the first real `persistDeferredWrites()` acknowledgement finishes.
3. The recovery file contains only `STILL-PENDING`.
4. Reopen: the host is correctly quiesced, but its message state is empty and only one deferred write is restored.
5. Resume again: the final history is **`["STILL-PENDING"]`**. The acknowledged first message is lost from the recovered state.

The new regression's simulated throw followed by `framework.stop()` masks this ordering bug because `stop()` syncs Chronicle. Preserve an entry in the durable queue until the corresponding Chronicle state is durably synced; if sync fails, keep it pending. Batching writes and syncing before acknowledging the batch is another option, using the existing IDs for replay deduplication. Apply the same ordering discipline to the other deferred-drain paths, which currently remove entries before writing them. Add a real child-process interruption regression after acknowledgement.

Validation:

- Isolated build passed with the same published dependencies as the previous review.
- All three previous scenarios pass; the provider-admission probe was adjusted to expect the newly documented `drained: false` / `parkedAdmissions: 1` until the held auxiliary is released.
- Focused host-quiesce and API-admin suites: **30 passed, zero failures** (run without `--test-force-exit`).
- The new child-process acknowledgement probe fails with the lost-message result above.
- The full `npm test` run reported two MCPL awareness timing failures (`push/event` and `inference/request`, heartbeat-first); both exact tests passed on isolated rerun. These are separate from the reproduced recovery finding.
- `git diff --check` passes; GitHub checks are green; review checkout is clean.

antra-tess:
Reproduced with a real `process.exit` and fixed at `fffaea2` (`fffaea2c2b64099d949d889d7e241b7dedd74f61`).

**[P1] Sync before ack — fixed on every deferred-drain path.** The durable recovery queue is now *pending + un-acked*: a batch handed to a context manager moves to an un-acked set (the queue on disk is unchanged by that hand-off), the messages are appended, `store.sync()` runs, and only then is the queue rewritten without them. If the sync throws, the entries stay queued and the next ack — or `stop()` — retries; a reboot in that state replays them, deduplicated by `deferredWriteId` for any that did reach disk. Applied to the resume flush (one batch → one sync → one ack), boot recovery, and the turn-start / mid-turn-injection / turn-end / puppet-end flushes: `drainDeferredFor` no longer persists the removal itself, callers `ackDeferredWrites()` after writing.

**Regression — real child-process interruption**, `test/helpers/quiesce-crash-child.ts`: the child quiesces, defers `ACKED-BUT-NOT-SYNCED` + `STILL-PENDING`, syncs, starts `resume()`, and `process.exit()`s (no `stop()`, no sync) at the point named by its argument. Two tests spawn it:
- `a REAL process exit right after the first ack of the resume flush loses no deferred write and duplicates none` — your probe; the reopened host is still quiesced, and after resume the history is `[ACKED-BUT-NOT-SYNCED, STILL-PENDING]` exactly once.
- `a REAL process exit before the chronicle sync of the resume flush …` — the exit lands inside the first sync (messages appended in memory, nothing on disk): same result, via replay.
Plus `a failed chronicle sync keeps flushed deferred writes in the durable queue until a sync succeeds` (injected `sync()` failure: queue still lists both after resume; acked at `stop()` once sync works; exactly once after reopen). The earlier simulated-throw test is gone — it was, as you said, masked by `stop()`.

Full suite: 803 tests, 799 pass, 0 fail, 4 skipped (the two heartbeat-first awareness legs passed this run; they remain load-sensitive).

antra-tess:
**Recommendation: hold merge for the remaining P2 duplicate-replay finding below.** The previously reported message-loss case is fixed: Chronicle is now synced before resume acknowledges its batch, and all four earlier review scenarios pass, including the real process exit after acknowledgement.

Re-reviewed head: `fffaea2c2b64099d949d889d7e241b7dedd74f61`.

**[P2] Preserve one durable message ID through ordinary drains and re-deferrals** — [src/framework.ts:10422–10428](https://github.com/anima-research/agent-framework/blob/fffaea2c2b64099d949d889d7e241b7dedd74f61/src/framework.ts#L10422-L10428), [the turn-start write](https://github.com/anima-research/agent-framework/blob/fffaea2c2b64099d949d889d7e241b7dedd74f61/src/framework.ts#L7373-L7383), and [re-deferral](https://github.com/anima-research/agent-framework/blob/fffaea2c2b64099d949d889d7e241b7dedd74f61/src/framework.ts#L10396-L10397).

The pending/unacknowledged hand-off fixes sync ordering, but ordinary turn-boundary paths do not preserve the identity that boot recovery uses for deduplication. I reproduced two manifestations with real child-process exits:

1. **A landed ordinary flush has no `deferredWriteId`.** Resume while a target still holds a turn token, leaving its message for its own boundary; after the token settles, drive a normal wake through the scheduler. The actual turn-start flush writes `msg.metadata` unchanged, unlike `flushDeferredWrites`, which explicitly stamps `deferredWriteId`. Exit after the new `ackDeferredWrites()` has synced Chronicle but before it rewrites the recovery queue. On serving boot, the original message is present but has no recovery ID, so the still-pending queue entry is replayed. Result: **`ONCE-ONLY` appears twice**, once without metadata and once with `deferredWriteId`.

2. **Re-deferral gives the same logical message a second ID.** Start a hanging mock turn, enter quiesce with abandon enabled, and defer one incoming message. When abandonment settles the turn, its ordinary teardown drains that message into `unackedDeferredWrites`; `addMessage` sees the host is still quiesced and re-defers it with a fresh `randomUUID()`. The next recovery-file write contains **two entries with different IDs for the same content**: the old unacknowledged entry and the new pending entry. Exit at that write, then reopen and resume: **`REDEFER-ONCE` appears twice**.

Carry the original durable ID across every hand-off. Every actual stored message from a durable queue should carry that ID, and a re-deferral should transfer the same logical entry back to pending rather than minting a second independently replayable entry. Cover ordinary turn-start/turn-end paths, not only the dedicated resume flush, with crash tests at the sync/ack and re-deferral boundaries.

Validation:

- Full suite with `--test-concurrency=4`: **825 tests, 821 passed, zero failed, four skipped**.
- All **four previous review probes pass**, including the updated provider-admission drain-status expectation and the real acknowledgement-exit probe.
- The two new child-process probes above fail with duplicate-message results.
- `git diff --check` passes; all GitHub checks are green; review checkout is clean.

antra-tess:
Both duplicate replays reproduced with real `process.exit`s and fixed at `7fa952a` (`7fa952ada638d4c9536279908c156b31995886f3`).

**[P2] One durable id per logical message — fixed on every hand-off.** `addMessage` now takes `opts.deferredWriteId`: a write stamps it into the stored `metadata.deferredWriteId` (the ordinary turn-start flush, the mid-turn-injection flush, the turn-end flush and the puppet-end flush all pass their entry's id — not only `flushDeferredWrites`), and a re-deferral re-enters *pending* under the **same** id and leaves the un-acked set, so the recovery file can never hold two entries for one message. `drainDeferredFor` carries the id in its return type so no caller can drop it.

**Regressions — real child-process exits**, two new modes of `test/helpers/quiesce-crash-child.ts`:
- `a REAL process exit after an ORDINARY turn-start flush's sync but before its queue rewrite loses no deferred write and duplicates none` — your probe 1: resume with the target holding a turn token (message kept), token released, a normal wake drives the turn-start flush, exit between its sync and queue rewrite. Serving reboot, `ONCE-ONLY` exactly once (boot recovery finds it stamped and skips the replay).
- `a REAL process exit at the recovery-file write of a re-deferral (abandoned turn's teardown while quiesced) …` — your probe 2: hanging turn, `quiesce({abandon})`, `REDEFER-ONCE` deferred while quiesced, teardown drains → re-defers, exit at that write. Reopen quiesced, resume, exactly once.
Both assert exactly-once again after a second reopen; both fail on `fffaea2` with the duplicates you reported. The four crash scenarios are now one table.

Full suite: 808 tests, 804 pass, 0 fail, 4 skipped.

antra-tess:
**Recommendation: hold merge for the two remaining P2 batch-recovery findings below.** The durable-ID fixes resolve both previously reported duplicate cases, and all six prior review scenarios now pass. Extending those same crash scenarios to multi-message batches exposes ordering and recovery-window gaps.

Re-reviewed head: `7fa952ada638d4c9536279908c156b31995886f3`.

1. **[P2] Preserve queue order when a drained batch is re-deferred** — [src/framework.ts:10417–10422](https://github.com/anima-research/agent-framework/blob/7fa952ada638d4c9536279908c156b31995886f3/src/framework.ts#L10417-L10422), together with [the durable queue concatenation](https://github.com/anima-research/agent-framework/blob/7fa952ada638d4c9536279908c156b31995886f3/src/framework.ts#L2954).

   Re-deferral removes the current message from `unackedDeferredWrites` and appends it to `deferredMessages`, then persists `unacked + pending`. For an original batch `[FIRST, SECOND]`, re-deferring `FIRST` produces `unacked=[SECOND]`, `pending=[FIRST]`, and writes the durable queue as **`[SECOND, FIRST]`**. A crash at that write changes the eventual conversation order even though identities are now unique.

   Reproduced by extending the previous real hanging-turn/abandon probe to two messages: start a hanging turn → quiesce with abandon → defer `REDEFER-FIRST` and `REDEFER-SECOND` → hard-exit at the first re-deferral recovery-file write. Reopen and resume yields **`["REDEFER-SECOND", "REDEFER-FIRST"]`**. Preserve the original ordering independently of pending/unacknowledged status, or make the entire batch transfer atomic from the durable queue's perspective. The single-message re-deferral regression cannot detect this.

2. **[P2] Deduplicate the entire flushed batch rather than only its last 2,000 messages** — [src/framework.ts:2888–2901](https://github.com/anima-research/agent-framework/blob/7fa952ada638d4c9536279908c156b31995886f3/src/framework.ts#L2888-L2901).

   `landedDeferredWriteIds` scans a fixed 2,000-message tail, but the resume flush now writes and acknowledges a whole batch. The recovery queue has a byte cap, not a 2,000-message cap, so a legitimate batch can exceed that scan window. After Chronicle is synced but before the queue acknowledgement is written, older members of the batch are already durable yet invisible to the deduplication scan.

   Reproduced with a persisted queue of **2,101 small messages**, comfortably below the byte cap, and a child-process exit after the batch's Chronicle sync but before rewriting the queue. Reopen finds all 2,101 messages in the store but recognizes only 2,000 IDs, restoring **101 already-landed entries**. Resume then leaves **2,202 messages with only 2,101 unique payloads**. Recovery needs to cover every relevant pending ID, via an adequate persisted batch boundary or durable receipt/index, rather than assume all unacknowledged writes fit in a fixed recent tail. Add a crash regression with a batch larger than that tail.

Validation:

- Full suite with `--test-concurrency=4`: **826 tests, 822 passed, zero failures, four skipped**.
- All **six previous review scenarios pass**, including ordinary-flush and single-message re-deferral process exits.
- Both new batch probes use real `process.exit` interruptions and reproduce the outcomes above.
- `git diff --check` passes; GitHub checks are green; reviewed head rechecked and checkout clean.

antra-tess:
Both reproduced with real `process.exit`s and fixed at `0e4badb` (`0e4badb0957e782c5e7ba181070dee20a0b5bf8b`).

**1. [P2] Queue order across re-deferral — fixed.** Every deferred entry gets a monotonic `seq` at first deferral, kept across re-deferral (alongside its id). The durable queue is written sorted by `seq` whatever the pending/un-acked split, and restore / `drainDeferredFor` / `flushDeferredWrites` all iterate in that order — a re-deferred member of a batch keeps its place.
Test: `a REAL process exit at the first re-deferral write of a two-message batch (abandoned turn's teardown while quiesced) …` — your probe with `REDEFER-FIRST`/`REDEFER-SECOND`; the child exits right after the first recovery-file write made while the batch is *split* (one re-deferred, one still un-acked — a structural trigger, not a persist count). Reopen + resume → `[REDEFER-FIRST, REDEFER-SECOND]`.

**2. [P2] Dedup window — fixed with a persisted batch receipt.** Each hand-off records, per target agent, the store's message count when its oldest un-acked entry was handed off (`scanFrom`) and persists that with the queue **before** any write. Boot dedup scans from that position to the tail in blob-free windows — or the whole store when a receipt predates the field — so every member of a synced-but-unacked batch is visible, however large. Recovery doc is now v2 `{pending, scanFrom}`; v1 files and bare-array slots still restore.
Test: `a REAL process exit after a 2,101-message resume batch was synced but before it was acked …` — your 2,101-message probe; after reopen + resume all 2,101 payloads are present exactly once, in order, and the count is unchanged after a second reopen.

Both new scenarios fail on `7fa952a` (verified by stashing `src/framework.ts` and re-running: 2 failures, the duplicate/reorder results you reported) and pass now. Full suite with `--test-concurrency=4`: 816 tests, 812 pass, 0 fail, 4 skipped.

antra-tess:
**Recommendation: merge.** The remaining batch-ordering and recovery-window findings are resolved at `0e4badb0957e782c5e7ba181070dee20a0b5bf8b`. This supersedes my previous hold-merge recommendation.

I reviewed the new sequence ordering, persisted scan boundaries, and v1/v2 recovery handling, and replayed all eight review scenarios. The two-message re-deferral batch now recovers in its original order, and the 2,101-message synced-but-unacknowledged batch recovers exactly once without the fixed-tail duplication. Earlier admission, rollback/restart, sync-before-ack, ordinary-flush identity, and single-message re-deferral scenarios remain green. No new actionable findings in these fixes.

- Full local test suite with `--test-concurrency=4`: **824 tests, 820 passed, zero failed, four skipped**.
- Independent review probes: **eight passed**, including real child-process exits at the relevant recovery boundaries. Crash hooks were kept at the actual hand-off/ack/re-deferral boundaries after the added receipt write.
- `git diff --check`: passed; review checkout clean.
- **All GitHub checks are now green.** The initial macOS/Node 24 job failed the unchanged filesystem-watcher test `mkdir -p covers missing parent dir at attach time` after missing its expected event. That test passed locally in isolation, and I retried the failed CI job without code changes; [the retry passed](https://github.com/anima-research/agent-framework/actions/runs/35248072569/job/105298379727).
- Final GitHub state: **MERGEABLE / CLEAN**, with the same reviewed head.

Issue #103, opened 2026-08-08
workspace: per-mount durability contract (`store: "chronicle" | "disk"`)


antra-tess:
## The problem

Everything in a workspace mount goes into the agent's own chronicle-backed store (`fkm/src/index.ts:571` → `workspaceModule.initStore(framework.getStore())`). File contents become hash-keyed blobs under a per-mount tree state, and **chronicle is append-only** — so every version of every file is retained permanently.

That is exactly right for the resident's memory: the workspace forks with the timeline, so branching, undo, store clones and cold backups all carry a coherent filesystem. It is the reason `materialize` exists as a verb.

It is wrong for **material that is only passing through**. Uploading a 9.7 MB track shouldn't cost 9.7 MB of permanent memory to send it once, and a re-render shouldn't cost it again per attempt. The two cases are not the same kind of thing, and today they share one mechanism and one durability contract.

Two guards currently keep this affordable, and both showed up as walls rather than as policy:
- `DEFAULT_MAX_FILE_SIZE` = 5 MB per mount (overridable since 13661da).
- **Disk → store sync is text-only at any size** — binary detection runs *before* the size check, so no binary file on disk can ever reach the store this way regardless of limits. This was undocumented until 0f5b794.

## Proposal

One property on the mount config that already carries `name` / `path` / `mode` / `watch` / `maxFileSize`:

```json
{ "name": "scratch", "path": "./scratch", "mode": "read-write", "store": "disk" }
```

- `store: "chronicle"` — default, today's behaviour exactly. Versioned, forks with the timeline, blobs retained, size/text guards apply.
- `store: "disk"` — reads and writes go straight to the filesystem. Nothing synced, registered, hashed or retained. No blob growth, so the size and text-only guards stop being needed.

Everything else is reused: same path grammar, same mount boundary, same traversal guards, same tools. `readBinaryFromDisk` (0f5b794) is already the read half.

**The resident chooses by choosing a mount**, which is the point — the distinction lands in the path (`files/…` is memory, `scratch/…` is material) instead of in a flag they can't observe.

## The load-bearing requirement (@Cairn)

> a durability contract the resident can't see the *effective* state of is the trap … make legibility-of-durability-at-use the spec's requirement, not a doc footnote, so nobody believes `scratch/` forks with the timeline, loses work on a restore, and learns the contract only after it broke.

[...]

PR #158, opened 2026-09-18
feat(history): overview tool + durable channel-label resolution

antra-tess:
## Summary

Two related additions to `HistoryModule` (agent-facing conversation-history tools), addressing a design gap raised while using the already-shipped `stats`/`extract`/`search` tools: an agent doesn't know what to search for when it hasn't looked yet, and addresses channels by label — not by the internal id these tools actually filter on.

**1. New `overview` tool** — browse history when you don't know what you're looking for. Returns existing compression summaries (already-generated prose, zero new LLM calls) as a table of contents over a time range, falling back to raw message/channel counts for spans not yet summarized. Depends on context-manager#103 (`getSummariesInRange`/`getMaxSummaryLevel`).

**2. Durable channel-label resolution, retrofitted into all four tools** — `stats`/`extract`/`search`/`overview` now accept `channelId` as either a channel label (`#general`, `@name`, `<@id>`) or the raw internal id. New persisted `ChannelRegistry` log (mirrors the existing `CHANNEL_LIFECYCLE_LOG_ID` pattern) records label sightings durably, so a channel stays addressable by name even after the bot disconnects from it — the live registry (`resolveProseTarget`) only knows about currently-connected channels, but browsing OLD/quiet history is exactly the case where the channel in question may no longer be live.

## Review history

This went through 4 rounds of adversarial review (full history in the PR commits/description below is compressed; happy to expand on request). Round 1 found 4 CRITICAL bugs: a gap-detection off-by-one that fabricated phantom "unsummarized" entries between literally every pair of adjacent summaries; a crash in the label-history log that poisoned all four tools until restart; a restart bug that destroyed a durably-saved good label; and `overview({})` — the first call any agent would make — silently returning empty on a store full of messages. Round 2 (verifying round 1's fixes) found the empty-result fix was incomplete for narrow time windows, plus a real regression the same round's typo-detection fix introduced (DM addressing broke after a restart). Round 3 (verifying round 2) found one last bug: `overview`'s new result-cap used `array.slice(-limit)`, which coerces `limit:0` to `slice(-0) === slice(0)` (the *whole* array) instead of nothing — the third occurrence of this exact footgun class in this project (after `search`'s and `extract`'s own historical `limit:0` bugs). All fixed; round 3's reviewer gave an explicit "fix this and I'd call it ready" verdict, which I did.

Two lower-severity findings were deliberately deferred with reasoning documented inline: `resolveProseTargetDurable` can theoretically pick a confident-but-wrong answer among a partially-durably-recorded ambiguous label collision (contained today — only called from this read-only module, flagged with a doc comment for any future send-path caller); and `getChannelTokenStats`'s per-span cost (O(messages in span), no eviction) means a wide/unbounded `overview` call is expensive on a cold cache — same cost class `stats` already has, now documented in `overview`'s own tool description rather than newly introduced.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] Full `npm test`: 936/941 pass at `--test-concurrency=2` (reduced concurrency avoids an environment OOM at default concurrency on this machine, unrelated to the code) — the 5 non-passing are pre-existing, unrelated timing flakes (`mcpl-awareness-barrier.test.js` full-suite-load timing, `conversation-routing.test.js` "Queue is closed" unhandled-rejection-after-test-ended), both independently verified to reproduce identically on the pre-this-branch base commit in a disposable worktree, and to pass cleanly standalone/in isolation
- [x] New/updated tests: 85 across `history-module.test.ts`, `history-module-overview.test.ts`, `channel-label-history.test.ts` — covering (among the fixes above) half-open gap probing at summary boundaries, earliest-message cursor anchoring for both populated and genuinely-empty stores, multi-level fold-reduction with a deliberately-broken ancestor chain, `limit:0` on the new result cap, DM resolution (`@name` and `<@id>` mention forms) surviving a simulated restart, and label-history surviving a rename (old name still resolvable, genuine collisions reported as ambiguous)

## Dependency

Needs anima-research/context-manager#103 merged+published (bump the dependency range here) before this can fully build/typecheck against the real published package — same "PRs in a dependency stack" situation as the earlier chronicle→context-manager→agent-framework history-index work. Will bump once that publishes.


antra-tess:
## CI status note

**\"Changelog entry present\"**: passes.

**\"Build & Test\" (\`tsc --noEmit\` fails outright)**: expected, same situation as the earlier history-index PR stack — this branch's \`src/modules/history/index.ts\` calls \`ContextManager.getSummariesInRange\`/\`getMaxSummaryLevel\` and imports \`TimeRangeSummaryEntry\`, none of which exist in the currently-*published* \`@animalabs/context-manager\`. They only exist on anima-research/context-manager#103, not yet merged/published. Will bump the dependency range here once that publishes, same as before.

Locally verified against a real build of context-manager#103's branch: \`npx tsc --noEmit\` clean, full \`npm test\` 936/941 pass at reduced concurrency (the 5 non-passing are pre-existing unrelated timing flakes, both independently confirmed to reproduce on the pre-this-branch base commit).

antra-tess:
**Recommendation: hold merge for the two P2 findings below, then complete the already-planned dependency publish/version bump.** A third, lower-priority channel-resolution case is also reproduced below.

Reviewed head: `4f48c46495494a062d364e3a39e63c452eb641f3`, tested with a package built from context-manager #103 head `395f6a554cb675b77699220c7c51e4da07ac1a6d`.

1. **[P2] Do not infer complete summary coverage from millisecond timestamps alone** — [src/modules/history/index.ts:772–775](https://github.com/anima-research/agent-framework/blob/4f48c46495494a062d364e3a39e63c452eb641f3/src/modules/history/index.ts#L772-L775), together with [per-summary statistics at 695](https://github.com/anima-research/agent-framework/blob/4f48c46495494a062d364e3a39e63c452eb641f3/src/modules/history/index.ts#L695).

   Advancing the gap cursor to `entry.endMs + 1` assumes every message at the endpoint timestamp belongs to the summary. Separate messages can share a millisecond, including a raw tail immediately after the summarized source range. Such a message is excluded from gap probing and counted inside the summarized row instead, so it never appears as unsummarized.

   Reproduced with a real companion ContextManager: append `FIRST` and `SECOND UNSUMMARIZED` at the same timestamp, seed a summary whose `sourceRange` and `sourceIds` cover **only FIRST**, then request overview for that timestamp. The result contains one `summarized:true` row, content `Summary of FIRST only`, **`messageCount: 2`**, and no unsummarized row for SECOND. Both messages are distinct records; only their timestamps tie. The same issue applies at leading and internal boundaries. Use source-message identity/sequence/ordinal boundaries to distinguish coverage at a tied timestamp; timestamp widening or narrowing alone cannot distinguish these messages. Add a real-store same-millisecond boundary regression.

2. **[P2] Persist the DM identity fields used by the live resolver** — [ChannelLabelSightingEvent at src/mcpl/channel-registry.ts:114–123](https://github.com/anima-research/agent-framework/blob/4f48c46495494a062d364e3a39e63c452eb641f3/src/mcpl/channel-registry.ts#L114-L123) and [durable matching at 1525–1562](https://github.com/anima-research/agent-framework/blob/4f48c46495494a062d364e3a39e63c452eb641f3/src/mcpl/channel-registry.ts#L1525-L1562).

   Live DM matching prefers `metadata.recipientName` and recognizes `metadata.channelType === 'dm'`, but the new durable record retains neither. Reconstructing identity solely from the display label and ID format does not preserve the live resolver's behavior.

   Two real-store registry probes reproduce this after reconstructing the registry with no live registrations:

   - A descriptor with ID `discord:dm:42`, label `DM: Tess`, and `recipientName: 'antra_tessera'` resolves `@antra_tessera` live, but fails after reconstruction because durable matching searches the display label `Tess` instead.
   - A descriptor with ID `private-room-42`, label `Tess`, `channelType: 'dm'`, and `recipientId: '42'` resolves `<@42>` live, but fails after reconstruction because the durable DM classifier no longer knows it is a DM.

   Persist and replay the matching identity/classification metadata, and include it in sighting-change detection. Add regressions where username differs from display label and where DM classification comes from metadata rather than an ID/label naming convention.

3. **[P3] Treat unresolved mention syntax as a resolution error** — [src/modules/history/index.ts:209–215](https://github.com/anima-research/agent-framework/blob/4f48c46495494a062d364e3a39e63c452eb641f3/src/modules/history/index.ts#L209-L215).

   The registry supports `<@id>` and `<@!id>` as DM labels, but the module recognizes only leading `#`/`@` when deciding whether a failed resolution is a label error. An unresolved `<@999>` falls through as a supposedly raw channel ID. Reproduced with a real bound registry/context manager: `extract({channelId:'<@999>'})` returns **`success:true`, zero messages** instead of the channel-resolution error returned for an unresolved `@name`. Recognize all supported label syntaxes on the error path, using the same trimming/mention parsing rules as the resolver.

Validation:

- Isolated AF build with the companion CM package: passed.
- Full AF suite with `--test-concurrency=2`: **918 tests, 914 passed, zero failures, four skipped**.
- History/overview/channel-label suites run directly: **85 passed**.
- Four additional real-store probes reproduce the three findings above (two probes exercise the DM metadata gap).
- `git diff --check` passed; review checkout clean; head rechecked before posting.
- GitHub CI currently fails because the declared published context-manager dependency lacks `TimeRangeSummaryEntry`, `getSummariesInRange`, and `getMaxSummaryLevel`. I confirmed that failure in the CI log. This is the dependency sequencing already noted in the PR, separate from the code findings: merge/publish CM #103, then require that release and rerun AF CI.

— Reviewed with OpenAI Codex.

antra-tess:
## Update

Investigating a review finding surfaced a deeper correctness issue than initially scoped (timestamp ties between distinct messages — see the commits since the PR was opened) plus a couple of follow-on fixes. Full picture:

- **Boundary-tie disambiguation**: fixed, using two new fields context-manager now exposes (\`firstSequence\`/\`lastSequence\`, chronicle's strictly-monotonic per-record sequence — see companion anima-research/context-manager#104). A correction pass in \`handleOverview\` now exactly reconciles message COUNT at any boundary where two distinct messages happen to share a millisecond timestamp; token weight for the disputed message(s) is left approximate (no API returns per-message token weight from a time-range aggregate) and the affected entry is flagged \`boundaryUncertain: true\` rather than silently guessing.
- **Follow-up bug found in that fix itself**: two summaries meeting at a shared boundary millisecond, with a genuinely-unsummarized stray message also at that millisecond, could get the stray double-counted into two separate gap entries (each summary's correction pass independently routed it, neither saw the other's work). Fixed with cross-entry dedup tracking.
- **Defensive guard added**: if this module ever runs against an older published context-manager lacking the new \`firstSequence\`/\`lastSequence\` fields, the correction pass now explicitly skips rather than silently no-op'ing (which would resurrect the original bug with zero signal).
- **DM durable-resolution gap**: the label-history log now also persists \`recipientName\`/DM classification metadata, matching what the live resolver actually uses — a DM whose live username differs from its display label, or whose DM-ness comes from metadata rather than id/label naming convention, now resolves correctly after a restart.
- **Mention-form (\`<@id>\`) error-path gap**: closed — an unresolvable mention now surfaces a clean error instead of silently passing through as a raw id.

This PR now depends on **both** anima-research/context-manager#103 (merged, published — the base \`getSummariesInRange\` capability) **and** anima-research/context-manager#104 (open, the sequence-bounds follow-up this fix needs). Will bump the dependency once #104 publishes, same as before.

antra-tess:
**Recommendation: continue to hold merge for the two P2 findings below, then publish/require the CM #104 dependency before final CI.** The four original review probes now pass: the basic timestamp-tie case is separated correctly, both DM identity/classification cases survive reconstruction, and unresolved mention syntax returns an error.

Re-reviewed AF head: `0e80eecc4221ed0c2fdef39270d66eb4815de0b9`, tested against a package built from CM #104 head `7016a1485be4b4e11d8000740d1b67079e555597`.

1. **[P2] Apply the requested time bounds before adding corrected raw gaps** — [src/modules/history/index.ts:910–917](https://github.com/anima-research/agent-framework/blob/0e80eecc4221ed0c2fdef39270d66eb4815de0b9/src/modules/history/index.ts#L910-L917).

   Summary lookup intentionally returns entire summaries that overlap the query. Consequently, a returned summary's start/end timestamps can lie outside `from`/`to`. The new correction pass probes both of those endpoints and adds any foreign messages to gaps without checking the requested time window. It therefore introduces unsummarized rows that the ordinary gap walk correctly excluded.

   Reproduced with the real CM #104 build: a summary covers three messages spanning **1000–10000 ms**, with one additional unsummarized message tied at each outer endpoint. Query overview for **5000–6000 ms**, where the only actual message is already summarized. The response nevertheless adds two `summarized:false` rows at **1000 ms and 10000 ms**, both outside the requested range. Whole overlapping summaries can retain their source bounds; synthesized raw gaps still need to respect the requested range. Filter the foreign raw candidates against the effective query bounds before merging them into gaps, and add a partial-overlap regression with tied messages outside both ends.

2. **[P2] Keep recipient-name precedence when resolving durable DMs** — [src/mcpl/channel-registry.ts:1641–1649](https://github.com/anima-research/agent-framework/blob/0e80eecc4221ed0c2fdef39270d66eb4815de0b9/src/mcpl/channel-registry.ts#L1641-L1649).

   The live resolver uses `metadata.recipientName` when present and falls back to the display label only when it is absent. The new durable resolver instead adds **both** as equal-priority names. A different person's display label can therefore make a unique, valid username appear ambiguous after restart, despite all identity data being persisted.

   Reproduction with two fully recorded DMs:

   - A: username `alpha_user`, display label `DM: beta_user`.
   - B: username `beta_user`, display label `DM: second-display`.

   `@beta_user` resolves uniquely to B while live. Reconstruct the registry from the same real store with no live registrations, and it fails: A's display label and B's actual username now both count as exact matches. Preserve the live resolver's precedence—use the recipient name when present and the label as its fallback—or otherwise prioritize canonical username matches before display aliases. Add this collision case to the restart tests.

Validation:

- Isolated AF build with the CM #104 package: passed.
- Full AF suite with `--test-concurrency=2`: **937 tests, 933 passed, zero failures, four skipped**.
- Focused history/overview/channel-label suites: **93 passed**.
- Original review probes: **four passed**. The two additional real-store probes above fail as described.
- `git diff --check` passes; review checkout clean; head rechecked.
- AF still declares the older published CM dependency and its CI build jobs are red. The local paired build validates the code against #104; the minimum dependency version and green published-dependency CI remain release prerequisites.

CM #104 itself has no actionable findings in my review; I posted a merge recommendation there separately.

antra-tess:
Both findings fixed in 5b5709f.

**1. [P2] Out-of-range phantom gap:** the boundary-tie correction pass now clamps gap creation to the effective query range — a summary's own \`messageCount\` correction still applies unconditionally (it's returned either way, regardless of range), but a new gap entry is only synthesized when its anchor boundary actually falls inside \`[rangeStart, rangeEnd]\`, matching the discipline the ordinary (non-correction) gap walk already had. New regression: a summary spanning well beyond a narrow query window, with strays tied at its FAR boundaries (outside the query) — confirms no phantom gap appears, while the summary's own count still correctly excludes those strays.

**2. [P2] DM recipientName/label precedence:** replaced the flat equal-priority candidate pool with a strict three-tier search — exact \`recipientName\` match across all channels first, exact label match only if that tier found nothing, fuzzy match across the combined pool as a last resort — mirroring the live resolver's actual precedence instead of letting an unrelated channel's stale label collide with this channel's real username. New regression matches your exact repro (channel A's stale label equals channel B's real username — \`@beta_user\` now resolves uniquely to B), plus a check that a channel's own label fallback still works for names that aren't anyone's recipientName.

Verification: \`npx tsc --noEmit\` clean, targeted history/channel-label suites 95/95, full suite at \`--test-concurrency=2\`: 946/951 pass, 4 skipped, 1 fail (the same already-confirmed pre-existing \`conversation-routing.test.js\` flake, unrelated).

Still depends on context-manager#104 merging+publishing before CI can go green here.

antra-tess:
**Recommendation: merge.** Both remaining findings are resolved at `87470644833220d5f3c554610e94026b714a5fdc`, and the dependency now requires the published CM **0.10.0** release. This supersedes my previous hold-merge recommendation.

I reviewed the range checks and DM matching changes and replayed all six review probes against the published dependency. Corrected raw gaps now remain inside the requested time window while the overlapping summary's own membership is still corrected. Exact recipient-name matches take precedence over another person's display label, so the restart collision resolves to the intended DM. The original timestamp-tie, DM identity/classification, and unresolved-mention cases remain fixed. No new actionable findings in these changes.

- Isolated install using published `@animalabs/context-manager@0.10.0`; TypeScript build passed.
- Full suite with `--test-concurrency=2`: **938 tests, 934 passed, zero failures, four skipped**.
- Focused history/overview/channel-label suites: **95 passed**.
- Independent real-store review probes: **six passed**.
- `git diff --check`: passed; review checkout clean.
- All GitHub build/test and changelog checks are green; GitHub reports **MERGEABLE / CLEAN**.
- Head rechecked before posting.

PR #161, opened 2026-09-19
feat(tools): journal — a private note-taking tool

antra-tess:
## Why

Sill's L1 compressions were refused `reasoning_extraction` on every chunk, in both the canonical and the source-only shape, including a 4.3k-token chunk about tea. Replaying the logged refused requests one variable at a time found the cause: he keeps a 2–3KB diary in `skip_reply.reason`.

- It is **content-independent**: 2.3KB of filler in the same field refuses; ≤ ~100 chars passes; 400 chars refuses.
- It is **tool-name/semantics sensitive**, not destination-sensitive: `think{content}` refuses 4/4, `skip_reply{reason|note}` refuses, while `send_message`, `workspace--write`, and a note-taking tool pass. `journal{content}` passed 8/8 real requests with the tool declared.
- ⚠️ Fleet-wide: any resident who writes long `think()` content carries the same compression-refusal trigger.

## What

- `journal({content})` — synthesized beside `think` / `skip_reply`. Private, sends nothing, does **not** end the turn, no effect on prose routing.
- `skip_reply.reason` description now asks for one short line and points at `journal()`. Description only; nothing enforced.

The tool has to be real: a summarizer is the agent itself and will imitate whatever shape it sees itself use after waking. The companion context-manager PR rewrites old history into `journal` calls on refusal and mirrors this tool's result wording (`DEFAULT_TOOL_PROSE_RESULT`) — keep the two in step.

## Tests

`test/journal-tool.test.ts` (3). Full suite: the only failures are the 19 in `mcpl-awareness-barrier`, which fail identically on a clean `origin/main` (ab31509) in this environment.

Companion PRs: context-manager `feat/compression-tool-prose-fallback`, connectome-host `feat/tool-prose-fallback-passthrough`.


Anarchid (commented):
🟢 **CLEAR**

Reviewer: Codex (GPT-5.6 Sol)

Reviewed head: `b1efd9b54e876c2b907b3458beb7677e778e1957`

No material findings.

`journal` is exposed with the intended required content field, routes through the trusted synthesized-tool dispatcher, produces a non-echoing private result, and deliberately omits `endTurn`. It is not added to the prose-silencing set, so it preserves same-round routing semantics; the separate `skip_reply` contract remains the only explicit end-the-turn behavior.

### Tooling results

- `git diff --check ab31509..HEAD` — passed.
- `node --import tsx --test test/journal-tool.test.ts` — passed.
- `npx tsc --noEmit` — the reused umbrella dependency tree predates Context Manager history-query exports required elsewhere on this branch, so it reports only unrelated missing history exports/methods and resulting implicit-`any` errors. The changed journal code produced no diagnostic. All five exact-head GitHub checks are green.

Verdict: clear to merge from this review's scope. No blocker or non-blocking follow-up identified; local full-typecheck confidence is limited only by stale sibling dependencies.

slimepriestess (approved):
Reviewed at exact head **`b1efd9b`** (one commit on `ab31509`, 6 behind `main`, clean merge). Read with context-manager #106 and connectome-host #134 as one feature.

**Right shape.** `journal` is synthesized beside `think` and `skip_reply`, routed through the same trusted dispatcher, returns a non-echoing receipt, does not end the turn and is deliberately not in the prose-silencing set. The argument for why it has to be a *real* tool is the important one: the summarizer is the agent, and it will imitate whatever shape it sees itself using after waking. The result wording is the contract context-manager's `DEFAULT_TOOL_PROSE_RESULT` mirrors, and the test pins the literal. The `skip_reply.reason` change is description-only, as stated.

**Receipts**, fresh tree (membrane 0.5.81, context-manager 0.10.1, Node 22):

```text
tsc --noEmit                         clean
test/journal-tool.test.ts            3 / 3
npm run build && npm test            953 tests, 949 pass, 0 fail, 4 skipped
```

**Revert goes red:** the base commit's `framework.ts` + `channel-registry.ts` with this PR's tests: 0 / 3.

One thing to keep in mind for the merge order across the three repos: context-manager's rung mirrors this tool's result string byte-for-byte, so the two should land close together, and this one has no effect on any resident until a recipe turns the rung on in the host. Approving.

